PR #17, opened 2023-09-22
Add gitops agent based sync engine

maciaszczykm:


agent/pkg/client/console.go
```diff
@@ -0,0 +1,37 @@
...
+	console "github.com/pluralsh/console-client-go"
+)
+
+type authedTransport struct {
+	token   string
+	wrapped http.RoundTripper
+}
+
+func (t *authedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("Authorization", "Token "+t.token)
```
maciaszczykm, line 16: Should this be `Token` or `Bearer`?

PR #33, opened 2023-09-26
Add operator helm chart

michaeljguarino:
just a 1 replica deployment with variables for deploy token and url


helm/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: deployment-operator
+description: creates a new instance of the plural deployment operator
+type: application
+version: 0.9.1
```
maciaszczykm, line 5: Should we update to 0.1.0 or something like that?

helm/values.yaml.liquid
```diff
@@ -0,0 +1,4 @@
+secrets:
+    deployToken: {{ configuration.deployToken }}
```
maciaszczykm, line 2: Not sure if this will have any impact, but indentation is 4 spaces instead of 2.

maciaszczykm, line 2: Indentation is 4 spaces instead of 2.

helm/Chart.yaml
michaeljguarino: We'll probably have a lot of prereleases for it, we'll eventually go to a 0.x version when it's stable

maciaszczykm (approved):

zreigz (approved):

floreks (approved):

PR #62, opened 2023-10-17
feat: cli-utils

zreigz:


pkg/sync/loop.go
```diff
@@ -4,14 +4,21 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/cli-utils/pkg/apply"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+)
+
+const (
+	// The field manager name for the ones agentk owns, see
+	// https://kubernetes.io/docs/reference/using-api/server-side-apply/#field-management
+	fieldManager = "application/apply-patch"
```
michaeljguarino, line 21: we should probably make this plural specific, eg `plrl/application/apply-patch`

pkg/sync/status.go
```diff
@@ -2,110 +2,111 @@ package sync
...
+		statsCollector.Handle(e)
+		switch e.Type {
+		case event.ApplyType:
+			gk := e.ApplyEvent.Identifier.GroupKind
+			name := e.ApplyEvent.Identifier.Name
+			if e.ApplyEvent.Error != nil {
+				msg := fmt.Sprintf("%s apply %s: %s\n", resourceIDToString(gk, name),
+					strings.ToLower(e.ApplyEvent.Status.String()), e.ApplyEvent.Error.Error())
+				if e.ApplyEvent.Status == event.ApplyFailed {
+					err = fmt.Errorf(msg)
```
michaeljguarino, line 33: these should be collected and passed as service errors in the updateComponents calls

pkg/sync/status.go
```diff
@@ -2,110 +2,111 @@ package sync
...
-			State:     toStatus(obj),
-			Group:     gvk.Group,
-			Kind:      gvk.Kind,
-			Version:   gvk.Version,
-			Namespace: obj.GetNamespace(),
-			Name:      obj.GetName(),
-			Synced:    true,
-		})
+func fromSyncResult(e event.StatusEvent) *console.ComponentAttributes {
+	return &console.ComponentAttributes{
```
michaeljguarino, line 73: I think this is missing the `State` field (which i think you can get from the khealth submodule of cli-utils)

michaeljguarino:
Mostly looks good, just some small details that seem to be missed in the updateStatus callback

pkg/sync/status.go
zreigz: this error is passed to service errors. It is passed here:
```
	if err := engine.updateStatus(id, components, errorAttributes("sync", err)); err != nil {
		log.Error(err, "Failed to update service status, ignoring for now")
	}
```

pkg/sync/status.go
zreigz: I have been working with this

floreks (approved):

PR #78, opened 2023-11-16
feat: introduce helm dependency update

zreigz:
This PR fixes helm deployment with chart dependency. 
The new functionality is added to manage helm repositories (add/update). 
Additionally to make it run in the container environment the helm repo cache is set up in the /temp directory where can store all necessary files.


michaeljguarino (approved):
this looks good to me, how thoroughly have we tested it before release?

zreigz:
> this looks good to me, how thoroughly have we tested it before release?

I have tested it quite well. There always can be some edge cases

PR #89, opened 2023-12-04
feat: More robust wss publisher interface

michaeljguarino:
This makes this code a bit more modular, and also should set us up well for listening for gate events.  Also covers preventing CRD deletion

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206097337636687


zreigz (approved):
works fine :+1:

PR #101, opened 2024-01-02
feat: add unit tests for helm and kustomize templates

zreigz:


michaeljguarino (commented):
What do these tests do functionally?  It doesn't seem like they're serving up a tarball but not entirely sure i'm reading it right

zreigz:
> What do these tests do functionally? It doesn't seem like they're serving up a tarball but not entirely sure i'm reading it right

they read from a predefined test directory with different templates. They execute the 'Render' function to test templating

michaeljguarino (approved):

zreigz:
@michaeljguarino please merge this PR because I am not authorized

michaeljguarino:
I changed the settings to now allow merges w/o all gates passing so you can merge in the future @zreigz

PR #105, opened 2024-01-10
feat: improve crd status

zreigz:
The agent should intelligently understand CRD status. This generic method searches for `status` and fetches the conditions. If the condition `Ready` is true then the object is in a `healthy` state.

I have tested it with CD and Cluster API objects


michaeljguarino (approved):

zreigz:
@michaeljguarino I can't merge this PR :(

michaeljguarino:
Really, I don't think the branch rule should prevent that (it doesn't require status checks anymore and you have write permissions)

PR #106, opened 2024-01-15
add lua script support

zreigz:
This PR add support for the Lua script. The script is stored in the `CustomHealth` object.
```
kind: CustomHealth
metadata:
  name: test
  namespace: default
spec:
  script: |
    healthStatus = {}
    if Obj.status ~= nil then
      local ready = "Ready"
      if statusConditionExists(Obj.status, ready) then
        healthStatus = {
          status="Progressing"
        }
        if isStatusConditionTrue(Obj.status, ready) then
          healthStatus = {
            status="Healthy"
          }
        end
      end
    end

```

For the input use `Obj` for the output:
```
healthStatus = {
            status="Healthy"
            message="hello"
}
```


pkg/lua/funcs.go
```diff
@@ -0,0 +1,35 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type Status struct {
+	Conditions []metav1.Condition
+}
+
+func isStatusConditionExist(s map[string]interface{}, condition string) bool {
```
michaeljguarino, line 13: perhaps rename to `statusConditionExists`

pkg/lua/lua.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	return output, nil
+
+}
+
+func GetFuncMap() template.FuncMap {
+	funcs := sprig.TxtFuncMap()
+	funcs["isStatusConditionTrue"] = isStatusConditionTrue
+	funcs["isStatusConditionExist"] = isStatusConditionExist
```
michaeljguarino, line 47: rename will need to propagate here

pkg/lua/funcs.go
zreigz: done

michaeljguarino:
Could we also rename the crd to something like `CustomHealthStatus`?  Feel like that might be a bit more clear for users

cmd/main.go
```diff
@@ -98,6 +105,21 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = (&controller.HealthConvertReconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+		Agent:  a,
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "HealthConvert")
+		os.Exit(1)
```
michaeljguarino, line 114: this is a bit different practice, but for now I think we shouldn't exit 1 here as there could be annoying races with the crds being created that we don't anticipate and it's not good for the agent to go down entirely.

If we could drop an error log instead that'd be safer for sure.

zreigz:
> CustomHealthStatus

The CRD also has a Status so it will look strange `CustomHealthStatusStatus` so I will change to `CustomHealth`

cmd/main.go
zreigz: done

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):

PR #107, opened 2024-01-15
feat: gate-operator

rauerhans:
Adds the job gates feature.

- `api/pipelines/v1alpha1` defines the kubernetes CRD `PipelineGate` types in group `pipelines.plural.sh`
- `pkg/client/pipelines.go` implements interaction with console API + parsing into CRDs
- `pkg/controller/pipelinegates` implements the "agent-reconciler" that fetches pipeline gates from the console, and. creates/updates the `PipelineGate` CRDs in the cluster to reflect the gates in the console.
- `controllers/pipelinegates` implements the cluster-side controller of the `PipelineGate` CRD

- `make manifests` to update the CRD definitions
run with 
- `go run cmd/* -console-url=https://console.gatetest.onplural.sh/gql -cluster-id=15909239-6fac-472d-934f-6c793933cd79 -leader-elect=false -processing-timeout=5m -enable-helm-dependency-update=false -local=true -zap-devel -zap-log-level 1`


charts/deployment-operator/crds/velero.io_backuprepositories.yaml
maciaszczykm, line 1: I think all Velero CRDs should not be there. It might be merge/rebase leftover.

pipelines.yaml
maciaszczykm, line 1: If these are still needed can you move them to some other location?

config/samples/pipelines_v1alpha1_pipelinegate.yaml
maciaszczykm, line 1: You can probably put the file I commented on before there.

config/rbac/role.yaml
```diff
@@ -20,13 +20,39 @@ rules:
...
   - deployments.plural.sh
   resources:
-  - healthconverts/finalizers
+  - customhealths/finalizers
   verbs:
   - update
 - apiGroups:
   - deployments.plural.sh
   resources:
-  - healthconverts/status
```
maciaszczykm, line 29: Should this change? To verify I would rerun all codegen targets from Makefile.

pkg/client/genericcache.go
maciaszczykm, line 1: I think on main branch there is `cache.go` which has basically same logic. It also uses generic. If these are the same then we can remove it.

pkg/controller/service/reconciler.go
```diff
@@ -354,7 +354,7 @@ func (s *ServiceReconciler) defaultInventoryObjTemplate(id string) (*unstructure
 	}, nil
 }
 
-func disableClientLimits(config *rest.Config) {
+func DisableClientLimits(config *rest.Config) {
```
maciaszczykm, line 357: Can you export it to some common/utils package since it seems to be used in multiple packages now?

internal/controller/restore_controller.go
maciaszczykm, line 1: This file should be removed. It doesn't exist on main.

michaeljguarino (commented):
Are the pipeline gate crds in the deployment agent chart yet?  We'll need that for it to be deployable (but might have been added in a prior pr).  Otherwise seems to be good to go

api/pipelines/v1alpha1/doc.go
maciaszczykm, line 1: I think this should go directly to `api/v1alpha1` next to other type definitions.

maciaszczykm: Same as all other files from this directory.

api/pipelines/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "pipelines.plural.sh", Version: "v1alpha1"}
```
maciaszczykm, line 29: Oh, this is in different directory as this has different group?

pkg/controller/pipelinegates/reconciler.go
```diff
@@ -0,0 +1,189 @@
...
+	gateQueue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+
+	f := utils.NewFactory(config)
+
+	cs, err := f.KubernetesClientSet()
+	if err != nil {
+		return nil, err
+	}
+
+	genClientset, err := generated.NewForConfig(config)
```
maciaszczykm, line 71: You could also use runtime client like we did in restore controller (`k8sClient`). Then perhaps we would need to generate informers etc.

pkg/controller/pipelinegates/reconciler.go
```diff
@@ -0,0 +1,189 @@
...
+	logger.Info("attempting to sync gate", "id", id)
+	gate, err := s.GateCache.Get(id)
+	if err != nil {
+		logger.Error(err, "failed to fetch gate: %s, ignoring for now")
+		return
+	}
+
+	defer func() {
+		if err != nil {
+			logger.Error(err, "failed to reconcile gate", "Name", gate.Name, "Id", gate.ID)
```
maciaszczykm, line 130: You can remove whole `defer` block. Error will be logged from the method above.

pkg/controller/pipelinegates/reconciler.go
```diff
@@ -0,0 +1,189 @@
...
+			Name: namespace,
+		},
+	}, metav1.CreateOptions{})
+
+	if err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+	}
+	return nil
```
maciaszczykm, line 181:
```suggestion
	if apierrors.IsAlreadyExists(err){
			return nil
	}
	return err
```

pkg/controller/restore/reconciler.go
```diff
@@ -103,7 +102,7 @@ func (s *RestoreReconciler) Reconcile(ctx context.Context, id string) (result re
 	logger.Info("attempting to sync restore", "id", id)
 	restore, err := s.RestoreCache.Get(id)
 	if err != nil {
-		fmt.Printf("failed to fetch restore: %s, ignoring for now", err)
+		logger.Error(err, "failed to fetch restore: %s, ignoring for now")
```
maciaszczykm, line 105:
```suggestion
		logger.Error(err, "failed to fetch restore, ignoring for now")
```

controllers/pipelinegates/pipelinegate_controller.go
```diff
@@ -0,0 +1,376 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	client "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/go-logr/logr"
+	console "github.com/pluralsh/console-client-go"
+	pipelinesv1alpha1 "github.com/pluralsh/deployment-operator/api/pipelines/v1alpha1"
+
+	//job "k8s.io/api/batch/v1"
```
maciaszczykm, line 37:
```suggestion

```

controllers/pipelinegates/pipelinegate_controller.go
```diff
@@ -0,0 +1,376 @@
...
+	if err := r.Get(ctx, req.NamespacedName, gate); err != nil {
+		if apierrs.IsNotFound(err) {
+			log.V(1).Info("PipelineGate CR not found - skipping.", "Namespace", gate.Namespace, "Name", gate.Name)
+			return ctrl.Result{}, nil
+		}
+		log.Error(err, "Unable to fetch PipelineGate.")
+		return ctrl.Result{}, err
+	}
+
+	// INITIAL STATE
```
maciaszczykm, line 73: It would be easier to read if handling for each state would be moved to separate functions.

controllers/pipelinegates/pipelinegate_controller.go
```diff
@@ -0,0 +1,376 @@
...
+		if err := r.Status().Update(ctx, gate); err != nil {
+			log.Error(err, "Failed to update PipelineGate status at CR")
+			return ctrl.Result{}, err
+		}
+		log.V(1).Info("Updated state of CR on first reconcile.", "Namespace", gate.Namespace, "Name", gate.Name, "ID", gate.Spec.ID, "SyncedState", gate.Spec.SyncedState, "LastSyncedAt", gate.Spec.LastSyncedAt, "State", gate.Status.State)
+		return ctrl.Result{}, nil
+	}
+
+	// PENDING
+	if gate.Status.State != nil && *gate.Status.State == pipelinesv1alpha1.GateState(console.GateStatePending) {
```
maciaszczykm, line 87: It would be nice to add methods on gate state to check that. Then conditions here would look simpler.

maciaszczykm: I.e. like [here](https://github.com/pluralsh/console/blob/0ac7b9a976cb2aea159037047e9e931f32044dc1/controller/api/v1alpha1/cluster_types.go#L389).

api/pipelines/v1alpha1/pipelinegate_types.go
```diff
@@ -0,0 +1,81 @@
...
+	Spec   PipelineGateSpec   `json:"spec,omitempty"`
+	Status PipelineGateStatus `json:"status,omitempty"`
+}
+
+// PipelineGateStatus defines the observed state of the PipelineGate
+type PipelineGateStatus struct {
+	State          *GateState              `json:"state,omitempty"`
+	LastReported   *GateState              `json:"lastReported,omitempty"`
+	LastReportedAt *metav1.Time            `json:"lastReportedAt,omitempty"`
+	JobRef         *console.NamespacedName `json:"jobRef,omitempty"`
```
maciaszczykm, line 52: We should use ObjectReference here instead.

controllers/pipelinegates/pipelinegate_controller.go
```diff
@@ -0,0 +1,376 @@
...
+			return ctrl.Result{}, nil
+		} else {
+			log.V(2).Info("Gate has a JobRef, checking Job status.", "Name", gate.Name, "ID", gate.Spec.ID, "State", *gate.Status.State, "jobRef", *gate.Status.JobRef)
+			job := r.generateJob(ctx, log, *gate.Spec.GateSpec.JobSpec, *gate.Status.JobRef)
+			if err := ctrl.SetControllerReference(gate, job, r.Scheme); err != nil {
+				log.Error(err, "Error setting ControllerReference for Job.")
+				return ctrl.Result{}, err
+			}
+			// reconcile job, creates a new one or gets the old one
+			reconciledJob, err := Job(ctx, r.Client, job, log)
```
maciaszczykm, line 135: We should create reconciler for jobs and change the gate status from it once job is finished. We could watch only for jobs that are connected with gates. Then this reconciler would be taking care only of gates (and creating job).

controllers/pipelinegates/pipelinegate_controller.go
```diff
@@ -0,0 +1,376 @@
...
+	for _, condition := range conditions {
+		if condition.Type == batchv1.JobFailed {
+			return true, &condition
+		}
+	}
+	return false, nil
+}
+
+func hasSucceeded(job *batchv1.Job) (bool, *batchv1.JobCondition) {
+	conditions := job.Status.Conditions
```
maciaszczykm, line 224: You can use `k8s.io/apimachinery/pkg/api/meta.IsStatusConditionPresentAndEqual` or other methods from this package. It can replace `hasFailed` and `hasSucceeded` functions. At the moment you only check if there is a condition on a job, but do not check its status (`true` or `false`).

controllers/pipelinegates/pipelinegate_controller.go
```diff
@@ -0,0 +1,376 @@
...
+			Name:      jobRef.Name,
+			Namespace: jobRef.Namespace,
+		},
+		Spec: jobSpec,
+	}
+}
+
+func CopyJobFields(from, to *batchv1.Job, log logr.Logger) bool {
+	requireUpdate := false
+	if !reflect.DeepEqual(to.Spec.Template.Spec.Volumes, from.Spec.Template.Spec.Volumes) {
```
maciaszczykm, line 289: Can we just copy whole spec.template.spec? Then we can just log that spec has changed. This seems to be a bit too detailed.

controllers/pipelinegates/pipelinegate_controller.go
rauerhans: that sounds like an antipattern, tbh, we're already modifying the state from two places

controllers/pipelinegates/pipelinegate_controller.go
rauerhans: I coulnd't find a method in that package or elsewhere that works with JobConditions, they seem to have their own struct types (for whatever reason), I will amend my methods though, to account for the `true` `false` check (it SHOULD always be `true` from what I can tell by looking at the job API but you never know

maciaszczykm:
@rauerhans I can see some of my review comments being hidden by GitHub, i.e. https://github.com/pluralsh/deployment-operator/pull/107#discussion_r1480162565. Make sure you solve this one in particular as it will create duplicates of CRDs otherwise.

api/pipelines/v1alpha1/pipelinegate_types.go
rauerhans: I would actually keep it this way, because that makes it easier to report it back to the console

config/rbac/role.yaml
rauerhans: idk I didnt' change it, I think that's the name now isn't it? maybe it wasn't updated before on main as this config dir isn't really used as a base for the chart

config/crd/bases/deployments.plural.sh_pipelinegates.yaml
michaeljguarino, line 1: why is this file registered twice?  (Oh is it a different api group?)

rauerhans:
> @rauerhans I can see some of my review comments being hidden by GitHub, i.e. [#107 (comment)](https://github.com/pluralsh/deployment-operator/pull/107#discussion_r1480162565). Make sure you solve this one in particular as it will create duplicates of CRDs otherwise.
> 
> Edit: [#107 (comment)](https://github.com/pluralsh/deployment-operator/pull/107#discussion_r1480182991) is also among hidden ones. Try to check them before merge.

Done

rauerhans:
I think this should be ready to go now, I'll be online later again, but go ahead if you want to merge it.

michaeljguarino (approved):

PR #109, opened 2024-01-16
feat: Add Helm labels

maciaszczykm:


pkg/manifests/template/helm.go
```diff
@@ -160,12 +170,60 @@ func (h *helm) templateHelm(conf *action.Configuration, name, namespace string,
 	if err != nil {
 		return nil, err
 	}
-	var manifests bytes.Buffer
-	_, err = fmt.Fprintln(&manifests, strings.TrimSpace(rel.Manifest))
+
+	manifests := strings.Split(rel.Manifest, "---")
```
michaeljguarino, line 174: I'm pretty sure there's a place further up in the code where we're handling these as kubernetes unstructured objects where it'd be more natural to just set the labels explicitly w/o the string parsing here.

michaeljguarino: Yeah perhaps just loop through the result of streamManifests in the `Render` function for the helm templater, which are all just unstructured objects which can be modified in-place

maciaszczykm: Oh, right. I didn't notice that stream manifests is using unstructured already.

maciaszczykm: Fixed.

michaeljguarino (approved):
this lgtm one small question but it's not too critical

pkg/manifests/template/helm.go
```diff
@@ -82,18 +93,57 @@ func (h *helm) Render(svc *console.ServiceDeploymentExtended, utilFactory util.F
...
 	}
 
-	out, err := h.templateHelm(config, svc.Name, svc.Namespace, values)
+	rel, err := h.templateHelm(config, svc.Name, svc.Namespace, values)
+	if err != nil {
+		return nil, err
+	}
+
+	var buffer bytes.Buffer
+	_, err = fmt.Fprintln(&buffer, strings.TrimSpace(rel.Manifest))
```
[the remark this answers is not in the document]
zreigz: we just write to the byte buffer

PR #114, opened 2024-01-25
refactor: Agent refactoring

maciaszczykm:


maciaszczykm:
Added restore controller here as well.

rauerhans (approved):

Makefile
```diff
@@ -16,7 +16,7 @@ $(LOCALBIN):
 
 ENVTEST_K8S_VERSION := 1.28.3
```
michaeljguarino, line 17: we should add a make target to copy in the velero crds to simplify that going forward

michaeljguarino (commented):
The one thing i'm not sure is duplicated fully is the websocket reconnects, I see the rejoin logic there, how has it been tested?

michaeljguarino (approved):
add makefile change and should be good to go

PR #118, opened 2024-02-09
feat: avoid fetching components for ServiceDeployment

floreks:
- Prepare specialized ServiceDeployment query for operator and minimize loaded number of fields.
- Get rid of the `components` sideload from ServiceDeployment query
- Omit components content when updating status for non dry-run services
- Read live resource status from K8S cluster when updating dry-run services
- Update `console-client-go` to latest version and use new client interface
- fix an issue with dry-run service components being always marked as synced


zreigz (approved):

pkg/manifests/template/utils.go
```diff
@@ -4,7 +4,7 @@ import (
 	console "github.com/pluralsh/console-client-go"
 )
 
-func configMap(svc *console.ServiceDeploymentExtended) map[string]string {
+func configMap(svc *console.GetServiceDeploymentForAgent_ServiceDeployment) map[string]string {
```
michaeljguarino, line 7: Could we have a more straightforward type for this?

floreks: Unfortunately, that's autogenerated by the `gqlgenc` package we use in `console-client-go`. The new, v2 client now generates struct names differently. It's `<queryName>_<returnType>`.

michaeljguarino (approved):

PR #120, opened 2024-02-12
feat: implement paginated polling agent-side

zreigz:
This PR implements two features:
 - paginated polling
 - service context templating


pkg/controller/service/reconciler.go
```diff
@@ -177,17 +177,25 @@ func (s *ServiceReconciler) ShutdownQueue() {
...
-	}
+	var after *string
+	var pageSize int64
+	pageSize = 100
+	hasNextPage := true
 
-	for _, svc := range svcs {
-		logger.Info("sending update for", "service", svc.ID)
-		s.SvcQueue.Add(svc.ID)
+	for hasNextPage {
```
michaeljguarino, line 186: We don't need to solve this right now, but feels like we could make a reusable poller interface with generics here.

michaeljguarino (approved):

PR #122, opened 2024-02-15
feat: Refactor gate operator

rauerhans:
https://linear.app/pluralsh/issue/ENG-1739/refactor-pipeline-job-reconciler

the names and namespaces now behave like this:

- `PipelineGate` CRs all go into `<operator namespace>-pipelines`, if this ns doesn't exist yet on the cluster, it is created by the operator.

- `PipelineGate` names are combination of the stripped name of the gate in the console (or wherever it was created) + the first 8 chars of the gate's ID

- If the job spec's pod template includes a namespace it will be created in that namespace, otherwise it will be created in the namespace of the pipeline gates

- Any one job owned by a particular `PipelineGate', will have a name equal to the name of the `PipelineGate` + 8 random uuid chars

e.g. a processed `PipelineGate`

```yaml
kind: PipelineGate
metadata:
  name: job-gate-1c31a49e
  namespace: plrl-deploy-operator-pipelines
status:
  jobRef:
    name: job-gate-1c31a49e-d97dc482
    namespace: plrl-deploy-operator-pipelines
  sha: YKWKEHUYT2XCGHW3YW6JZBI6K5M577V37XYKK56F7VVJCWYCOBGA====
  state: OPEN
spec:
  gateSpec:
    job:
      template:
        metadata: {}
        spec:
          containers:
            - args:
                - /bin/sh
                - '-c'
                - echo "Hello, World!"
                - echo $TEST_ENV_VAR
              env:
                - name: TEST_ENV_VAR
                  value: pipeline
              image: alpine:3.7
              name: job-gate-0
              resources: {}
          restartPolicy: OnFailure
  id: 1c31a49e-14cf-4cc1-b6c5-d4828d034be1
  name: job-gate
  type: JOB
```


internal/controller/pipelinegate_controller.go
```diff
@@ -53,59 +56,105 @@ type PipelineGateReconciler struct {
...
+	cachedGate, err := r.GateCache.Get(crGate.Spec.ID)
+	if err != nil {
+		log.Error(err, "Unable to fetch PipelineGate from cache, this gate probably doesn't exist in the console.")
+		return ctrl.Result{}, err
+	}
+
+	scope, _ := NewPipelineGateScope(ctx, r.Client, crGate)
+	defer func() {
+		attrs, _ := crGate.Status.GateUpdateAttributes()
+		sha, _ := utils.HashObject(attrs)
```
maciaszczykm, line 81: I think it may be risky to skip these errors (even if the chance of error is low). Setting wrong, i.e. empty, SHA can cause issues later.

api/v1alpha1/pipelinegate_types.go
```diff
@@ -99,6 +103,74 @@ func (pgs *PipelineGateStatus) HasJobRef() bool {
...
+	return &updateAttributes, nil
+}
+
+func (pgs *PipelineGateStatus) SetSHA(sha string) *PipelineGateStatus {
+	pgs.SHA = &sha
+	return pgs
+}
+
+func (pg *PipelineGate) CreateNewJobName() string {
+	jobName := fmt.Sprintf("%s-%s", pg.Name, uuid.New().String()[:8])
```
zreigz, line 166: maybe it would be better to use this:
```
import  "k8s.io/apimachinery/pkg/util/rand"


jobName := fmt.Sprintf("%s-%s", pg.Name, rand.String(8))
```

internal/controller/pipelinegate_controller.go
```diff
@@ -124,19 +173,7 @@ func (r *PipelineGateReconciler) reconcilePendingGate(ctx context.Context, gate
...
-		// try to update gate at console
-		attributeState := console.GateStatePending
-		updateAttributes := console.GateUpdateAttributes{State: &attributeState, Status: &console.GateStatusAttributes{JobRef: gate.Status.JobRef}}
-		if err := r.ConsoleClient.UpdateGate(gate.Spec.ID, updateAttributes); err != nil {
-			log.Error(err, "Failed to update PipelineGate status to console")
-		}
 
-		log.V(1).Info("Created new job for gate and updated status.", "Name", gate.Name, "ID", gate.Spec.ID, "State", *gate.Status.State, "jobRef", *gate.Status.JobRef)
 		return ctrl.Result{}, nil
 	} else {
```
michaeljguarino, line 178: I think this else block can actually be removed since you always return in the above clause (and can save some indentation as a result)

internal/controller/pipelinegate_controller.go
```diff
@@ -152,64 +189,37 @@ func (r *PipelineGateReconciler) reconcilePendingGate(ctx context.Context, gate
...
+			// I don't think a guarantee for aborting a job is possible, unless we change the console api to allow for it
+			// try to kill the job
+			err = killJob(ctx, r.Client, job)
+			// even if the killing of the job fails, we better update the gate status to closed asap, so we don't report a gate CR transition from pending to closed
+			gate.Status.SetState(console.GateStateClosed)
+			if err != nil {
+				log.Error(err, "Failed to kill Job.", "JobName", job.Name, "JobNamespace", job.Namespace)
+				return ctrl.Result{}, err
+			}
+			log.V(1).Info("Job aborted.", "JobName", job.Name, "JobNamespace", job.Namespace)
```
michaeljguarino, line 203: similar here, just `return ctrl.Result{}, err` and reduce the need for indentation of the following

michaeljguarino (commented):
Overall this is a bunch cleaner, small style nits and then we can approve and merge it.

rauerhans:
resolved all comments

michaeljguarino (approved):

PR #129, opened 2024-02-20
feat: Add more sprig funcs

maciaszczykm:


michaeljguarino (approved):
yea all these are v helpful

PR #132, opened 2024-02-23
feat: pause agent during cluster restore

zreigz:


pkg/controller/service/reconciler.go
```diff
@@ -367,3 +374,8 @@ func (s *ServiceReconciler) GetLuaScript() string {
 func (s *ServiceReconciler) SetLuaScript(script string) {
 	s.LuaScript = script
 }
+
+func (s *ServiceReconciler) isClusterRestore(ctx context.Context) bool {
+	_, err := s.Clientset.CoreV1().ConfigMaps(s.RestoreNamespace).Get(ctx, RestoreConfigMapName, metav1.GetOptions{})
```
michaeljguarino, line 379: could we also have some sort of timeout (1h?) on this configmap in the (unlikely) event that the controller can't pick up the completion of the restore process

zreigz: the timeout is managed in the velero restore controller. https://github.com/pluralsh/deployment-operator/pull/132/files#diff-96c112a527874ccdca45af904d9d83aeff57929f0cfb920909e889198c2377a0R70

zreigz: done

michaeljguarino (approved):

PR #133, opened 2024-02-24
fix: Don't sideload jobs in pipeline gate queries

michaeljguarino:
This is both inefficient, and brittle to cases when the job is pruned on reruns.  Exposed the gate status in the api so we can do it properly (didn't realize it was never queryable).

### DIFF
--- a/pkg/client/pipelines.go
+++ b/pkg/client/pipelines.go
@@ -216,20 +216,23 @@ func IsJobType(pgFragment *console.PipelineGateFragment) bool {
 }
 
 func HasJobRef(pgFragment *console.PipelineGateFragment) bool {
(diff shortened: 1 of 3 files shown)


pkg/client/pipelines.go
```diff
@@ -216,20 +216,23 @@ func IsJobType(pgFragment *console.PipelineGateFragment) bool {
...
-			Name:      fragment.Job.Metadata.Name,
-			Namespace: *fragment.Job.Metadata.Namespace,
-		}
-	} else if fragment.Job != nil {
+	if fragment.Status != nil && fragment.Status.JobRef != nil {
 		jobRef = &console.NamespacedName{
-			Name:      fragment.Job.Metadata.Name,
-			Namespace: "",
+			Name:      fragment.Status.JobRef.Namespace,
+			Namespace: fragment.Status.JobRef.Name,
```
maciaszczykm, line 235: @michaeljguarino it looks like name and namespace are mixed here.

PR #135, opened 2024-02-27
feat: refactor gate operator

zreigz:


pkg/controller/pipelinegates/reconciler.go
```diff
@@ -148,47 +140,37 @@ func (s *GateReconciler) Reconcile(ctx context.Context, id string) (result recon
...
-	return
-}
+	logger.V(1).Info("Gate exists.", "Namespace", gateCR.Namespace, "Name", gateCR.Name, "ID", gateCR.Spec.ID)
+	scope, err := pgctrl.NewPipelineGateScope(ctx, s.K8sClient, currentGate)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	currentGate.Spec = gateCR.Spec
+	err = scope.PatchObject()
```
michaeljguarino, line 167: Why do we still need to patch the object here?

Also I don't believe you should be modifying the console gate spec from the agent, just the status (eg the job it points to and it's status entry)

zreigz: yes, it's not needed. We can only create the CRD here

zreigz: done

michaeljguarino:
The one thing I don't think we've implemented into the reconciler yet as well, is we should compute a sha for the updates against the console api (not for the job) and ignore mutation calls if that sha doesn't change (that will prevent us from initiating too many write requests upstream during the reconcile loop)

michaeljguarino:
> @michaeljguarino This job is generated from the part of API PipelineGateFragment (JobSpecFragment) so basically, it checks the API changes. The write operations are only performed during status changes

Yea but can't the reconciler trigger for all sorts of underlying reasons besides the job changing (pod state changes, just rerunning for random reasons)?

zreigz:
@michaeljguarino I've applied your comments PTAL

internal/controller/pipelinegate_controller.go
```diff
@@ -237,15 +237,20 @@ func (r *PipelineGateReconciler) updateJob(ctx context.Context, reconciledJob *b
 	return nil
 }
 
-func (r *PipelineGateReconciler) updateConsoleGate(gate *v1alpha1.PipelineGate) error {
-	updateAttrs, err := gate.Status.GateUpdateAttributes()
+func (r *PipelineGateReconciler) updateConsoleGate(gate *v1alpha1.PipelineGate, cachedGate *console.PipelineGateFragment) error {
+	sha, err := utils.HashObject(gateUpdateAttributes(cachedGate))
```
michaeljguarino, line 241: why not use `gate.Status.GateUpdateAttributes()` here to avoid the possibility of drift between the impl of both those functions?

zreigz: gate Status reflects the current state (rather final state because this method is called in defer function after all operations) and I have to compare it with the API state (cachedGate) to decide if I should save new API attributes.

michaeljguarino: Ah ok as long as it works that's fine w/ me.

michaeljguarino (approved):

PR #138, opened 2024-03-01
feat: don't set job's ttl

zreigz:
keep a succeeded/failed jobs

### DIFF
--- a/pkg/client/pipelines.go
+++ b/pkg/client/pipelines.go
@@ -14,6 +14,8 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
+const twentyFourHours = int32(86400)
+
 func (c *client) UpdateGate(id string, attributes console.GateUpdateAttributes) error {
 	_, err := c.consoleClient.UpdateGate(c.ctx, id, attributes)
 	return err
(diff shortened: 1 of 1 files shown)


pkg/client/pipelines.go
```diff
@@ -132,9 +132,7 @@ func JobSpecFromJobSpecFragment(gateName string, jsFragment *console.JobSpecFrag
 		gateNameAnnotationKey := v1alpha1.GroupVersion.Group + "/gatename"
 		jobSpec.Template.ObjectMeta.Annotations[gateNameAnnotationKey] = gateName
 	}
-	ttlSeconds := int32(7200)
```
michaeljguarino, line 135: We actually do want a ttl though, it just needs to be long enough that logs are still usable

michaeljguarino (approved):

pkg/client/pipelines.go
zreigz: maybe 24h, WDYT?

michaeljguarino: Yea that seems sane to me

pkg/client/pipelines.go
```diff
@@ -132,7 +133,8 @@ func JobSpecFromJobSpecFragment(gateName string, jsFragment *console.JobSpecFrag
 		gateNameAnnotationKey := v1alpha1.GroupVersion.Group + "/gatename"
 		jobSpec.Template.ObjectMeta.Annotations[gateNameAnnotationKey] = gateName
 	}
-	
+	jobSpec.TTLSecondsAfterFinished = lo.ToPtr(int32(86400))
```
michaeljguarino, line 136: could we make the int a const with a descriptive name so its less of a magic number

pkg/client/pipelines.go
zreigz: done

pkg/client/pipelines.go
zreigz: done

michaeljguarino (approved):

PR #143, opened 2024-03-13
feat: add CRD register controller

zreigz:
Dynamically register backup/restore controllers when it detects installed Velero CRDs


michaeljguarino (approved):

charts/deployment-operator/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: deployment-operator
 description: creates a new instance of the plural deployment operator
-appVersion: 0.4.13
-version: 0.4.13
+appVersion: 0.4.14
```
michaeljguarino, line 4: I don't believe this tag has been created yet has it?

michaeljguarino (changes requested):

charts/deployment-operator/Chart.yaml
zreigz: I've changed it. Should I reverse this change?

zreigz:
@michaeljguarino PTAL, I've reverted the chart changes

PR #144, opened 2024-03-14
feat: add constraint controller

zreigz:


internal/controller/constraint_controller.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+const missingLabelError = "missing label while attempting to map a constraint status resource"
+
+type ConstraintReconciler struct {
+	client.Client
+	Scheme        *runtime.Scheme
+	ConsoleClient consoleclient.Client
+	Reader        client.Reader
+	Constrains    map[string]*console.PolicyConstraintAttributes
```
michaeljguarino, line 32: I think this should be spelled `Constraints` everywhere

michaeljguarino (commented):
Is there a constraint pod status object for every instance of a constraint template?

zreigz:
> Is there a constraint pod status object for every instance of a constraint template?

generally, it's for every constraint object

internal/controller/constraint_controller.go
zreigz: done

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):
This lgtm but looks like there's a go mod conflict

PR #162, opened 2024-04-08
fix: respect Service.SyncConfig.NamespaceMetadata

zreigz:
The agent adds labels/annotations to namespaces from `Service.SyncConfig.NamespaceMetadata`

### DIFF
--- a/pkg/controller/namespaces/reconciler.go
+++ b/pkg/controller/namespaces/reconciler.go
@@ -3,6 +3,7 @@ package namespaces
 import (
 	"context"
 	"fmt"
+	"maps"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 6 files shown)


pkg/controller/namespaces/reconciler.go
```diff
@@ -123,6 +124,11 @@ func (n *NamespaceReconciler) UpsertNamespace(ctx context.Context, fragment *con
 	if fragment.Annotations != nil {
 		annotations = convertMap(fragment.Annotations)
 	}
+	if fragment.Service != nil && fragment.Service.SyncConfig != nil && fragment.Service.SyncConfig.NamespaceMetadata != nil {
```
michaeljguarino, line 127: This isn't actually supposed to be on the namespace reconciler, it's meant for the service reconciler, where we upsert a namespace (although we can add it here too just a bit duplicative)

michaeljguarino: Also we should respect `createNamespace` there too (if it's false)

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):

PR #163, opened 2024-04-09
feat: helm template with --dry-run=server

zreigz:


pkg/manifests/template/cache.go
```diff
@@ -1,27 +1,36 @@
...
 
 import (
+	"sync"
+
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 var (
 	namespacedCache = &gvkCache{cache: map[schema.GroupVersionKind]bool{}}
+	mapMutex        = sync.RWMutex{}
```
michaeljguarino, line 11: there's also a concurrent map we can use that might do this more efficiently

michaeljguarino: I think it's used in the other caches, i didn't even think about the need for it here

michaeljguarino (approved):

PR #186, opened 2024-05-10
feat: add helm capabilities API versions

zreigz:
Fetch API versions every 5min. Use those versions for helm templating

### DIFF
--- a/pkg/controller/service/reconciler.go
+++ b/pkg/controller/service/reconciler.go
@@ -69,7 +69,8 @@ type ServiceReconciler struct {
 	pinger          *ping.Pinger
 }
 
-func NewServiceReconciler(consoleClient client.Client, config *rest.Config, refresh time.Duration, restoreNamespace string) (*ServiceReconciler, error) {
+func NewServiceReconciler(ctx context.Context, consoleClient client.Client, config *rest.Config, refresh time.Duration, restoreNamespace string) (*ServiceReconciler, error) {
+	logger := log.FromContext(ctx)
 	utils.DisableClientLimits(config)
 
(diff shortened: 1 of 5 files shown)


zreigz:
@michaeljguarino PTAL

pkg/controller/service/reconciler.go
```diff
@@ -104,6 +106,13 @@ func NewServiceReconciler(consoleClient client.Client, config *rest.Config, refr
 		return nil, err
 	}
 
+	go func() {
+		//nolint:all
+		_ = wait.PollImmediateInfinite(time.Minute*5, func() (done bool, err error) {
+			return false, CapabilitiesAPIVersions(discoveryClient)
```
michaeljguarino, line 112: if this returns err does the poller stop?

pkg/controller/service/reconciler.go
```diff
@@ -104,6 +106,13 @@ func NewServiceReconciler(consoleClient client.Client, config *rest.Config, refr
 		return nil, err
 	}
 
+	go func() {
```
michaeljguarino, line 109: I think we should make one synchronous call prior to initializing the service reconciler to eliminate possible race-condition bugs

pkg/controller/service/reconciler.go
zreigz: yes, but I've changed it to print the error log

pkg/controller/service/reconciler.go
zreigz: done

michaeljguarino (approved):

PR #194, opened 2024-05-23
feat(harness): support approval flow and more

floreks:
```
TBD
```


michaeljguarino (approved):
Tested a good bit and this seems to all work now (with a few minor fixes of my own)

PR #196, opened 2024-05-26
feat: Properly handle all controller errors

michaeljguarino:
It looks like there might not be error handling for issues that happen pre execution, this will solve for that a bit better in lieu of a more robust way of reporting run completion.


zreigz (approved):

pkg/harness/controller/controller.go
```diff
@@ -128,6 +137,14 @@ func (in *stackRunController) toExecutable(ctx context.Context, step *gqlclient.
 	)
 }
 
+func (in *stackRunController) workdir() string {
```
floreks, line 140: Working dir is a harness app argument. It is used in a couple more places, i.e. to prepare the environment, download and unpack tarball, read/save terraform related files such as plan/state. It would be best if job controller could read and configure it during job creation.

pkg/harness/controller/controller.go
```diff
@@ -78,6 +79,14 @@ func (in *stackRunController) Start(ctx context.Context) (retErr error) {
 	return in.postStart(retErr)
 }
 
+func (in *stackRunController) Finish(stackRunErr error) error {
```
floreks, line 82: It could also be fixed by wrapping all early returns from the `Start` function with `postStart`

pkg/harness/controller/executor.go
```diff
@@ -125,14 +112,6 @@ func (in *executor) preRun(id string) {
...
 
-func (in *executor) postRun(id string, err error) error {
-	if in.postRunFunc != nil {
-		in.postRunFunc(id, err)
-	}
-
-	return err
-}
-
 func (in *executor) dequeue(executable exec.Executable) (empty bool) {
```
floreks, line 115: This is probably unused now

pkg/harness/controller/executor.go
```diff
@@ -51,32 +51,15 @@ func (in *executor) ordered(ctx context.Context) {
...
-	}()
-
 	// Read executables and run them in order
 	go func() {
-		for {
-			// Get executable from the queue
-			executable := <-in.ch
-
-			// Run the executable and wait for it to finish
+		for _, executable := range in.startQueue {
```
floreks, line 56: oh yeah, that's better. Good catch.

pkg/harness/controller/controller.go
michaeljguarino: The idea behind this workdir is it only controls the directory under which commands are executed.  The case is:

* sync in the entire terraform folder (with adjacent modules)
* execute under some subfolder

pkg/harness/controller/controller.go
michaeljguarino: I feel like handling this a level up is less brittle to devs forgetting to add the postStart wrapper

pkg/harness/controller/executor.go
michaeljguarino: oh yeah true

pkg/controller/stacks/job.go
```diff
@@ -205,12 +206,18 @@ func (r *StackReconciler) getDefaultContainerImage(run *console.StackRunFragment
...
-	return []string{
+func (r *StackReconciler) getDefaultContainerArgs(run *console.StackRunFragment) []string {
+	args := []string{
 		fmt.Sprintf("--console-url=%s", r.ConsoleURL),
 		fmt.Sprintf("--console-token=%s", r.DeployToken),
-		fmt.Sprintf("--stack-run-id=%s", runID),
+		fmt.Sprintf("--stack-run-id=%s", run.ID),
 	}
+
+	if run.Workdir != nil && len(*run.Workdir) > 0 {
```
michaeljguarino, line 216: We shouldn't do this like i said, it's actually intended to have the workdir be a subdirectory of the overall dir

pkg/harness/controller/controller_types.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/pluralsh/deployment-operator/internal/helpers"
 	console "github.com/pluralsh/deployment-operator/pkg/client"
 	"github.com/pluralsh/deployment-operator/pkg/harness/sink"
-	"github.com/pluralsh/deployment-operator/pkg/harness/stackrun"
+	v12 "github.com/pluralsh/deployment-operator/pkg/harness/stackrun/v1"
```
michaeljguarino, line 11: maybe `stackrunv1` instead of `v12`?

floreks: oh, that must have been automatically updated by my IDE during refactoring

pkg/harness/controller/controller.go
floreks: Restored

pkg/harness/controller/controller.go
floreks: Done. Tested and seems to be working fine.

pkg/controller/stacks/job.go
floreks: Moved to harness logic

PR #198, opened 2024-05-27
fix: Fix filepath setup

michaeljguarino:
With the new mountPath spec, you can assume users specify a full/correct path in harness

### DIFF
--- a/pkg/harness/environment/environment.go
+++ b/pkg/harness/environment/environment.go
@@ -1,8 +1,6 @@
 package environment
 
 import (
-	"path"
-
 	"k8s.io/klog/v2"
 
 	"github.com/pluralsh/deployment-operator/internal/helpers"
@@ -38,7 +36,7 @@ func (in *environment) prepareFiles() error {
 	}
 
 	for _, fragment := range in.stackRun.Files {
-		destination := path.Join(in.filesDir, fragment.Path)
+		destination := fragment.Path
 		if err := helpers.File().Create(destination, fragment.Content); err != nil {
 			klog.ErrorS(err, "failed preparing files", "path", destination)
 			return err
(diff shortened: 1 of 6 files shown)


pkg/harness/tool/terraform/modifier.go
```diff
@@ -34,6 +33,10 @@ func NewPlanModifier(planFileName string) *PlanModifier {
 }
 
 func (in *ApplyModifier) Args(args []string) []string {
+	if !lo.Contains(args, "apply") {
+		return args
+	}
+
 	if !helpers.IsExists(path.Join(in.dir, in.planFileName)) ||
 		// This is to avoid doubling plan file arg if API already adds it
 		algorithms.Index(args, func(a string) bool {
```
floreks, line 42: This could also be replaced with Contains func

floreks (approved):

pkg/harness/environment/environment.go
```diff
@@ -38,7 +36,7 @@ func (in *environment) prepareFiles() error {
 	}
 
 	for _, fragment := range in.stackRun.Files {
-		destination := path.Join(in.filesDir, fragment.Path)
```
floreks, line 41: IIUC it will require user to provide full path including harness working dir, i.e. `/plural/tf-test/...`. Isn't it more prone to errors? Any other root level dir than `/plural` will not work currently.

michaeljguarino: You should be able to specify `.` as the mount path, and then it would be something like `./creds.json`.  I do think we should probably make some other folders writable though i just can't think of which at the moment (also think there probably will need to be a home folder for the user).

The problem I faced though was I didn't know the path at config-time at first, so it was a bit tricky to set things like GOOGLE_APPLICATION_CREDENTIALS properly, if you have full understanding of the path that's more straightforward.

PR #199, opened 2024-05-28
feat: add job timeout

zreigz:
Add time out for stack run job reconciliation. After timeout mark the stack run with an error state.

### DIFF
--- a/internal/controller/stackrunjob_controller.go
+++ b/internal/controller/stackrunjob_controller.go
@@ -27,7 +27,6 @@ import (
 	"github.com/pluralsh/deployment-operator/pkg/client"
 	"github.com/pluralsh/deployment-operator/pkg/controller/stacks"
 	"github.com/pluralsh/polly/algorithms"
-	"github.com/samber/lo"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


internal/controller/stackrunjob_controller.go
```diff
@@ -42,6 +43,7 @@ import (
 )
 
 const jobSelector = "stackrun.deployments.plural.sh"
+const jobTimout = time.Minute * 10
```
michaeljguarino, line 46: Think this should be 30 or 60 minutes (could have a weird scheduling edge case that takes 10 minutes for benign reasons I think)

internal/controller/stackrunjob_controller.go
```diff
@@ -72,15 +74,24 @@ func (r *StackRunJobReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	// Update step statuses, i.e., when stack run was successful or failed.
 	for _, step := range stackRun.Steps {
```
michaeljguarino, line 75: Oh we should actually not do these updates here (let the harness fully own them and it isn't going to impact anything on the api if they remain pending)

internal/controller/stackrunjob_controller.go
```diff
@@ -72,15 +74,24 @@ func (r *StackRunJobReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				return ctrl.Result{}, err
+			}
 		}
 	}
 
 	// Exit if stack run is not in running state (run status already updated),
 	// or if the job is still running (harness controls run status).
 	if stackRun.Status != console.StackStatusRunning || job.Status.CompletionTime.IsZero() {
-		return ctrl.Result{}, nil
+		if isActiveJobTimout(stackRun.Status, job) {
```
michaeljguarino, line 86: I think we should also manually delete the job here, or perhaps kill its pod.  I'm not sure if the ttl on the job will pick up

internal/controller/stackrunjob_controller.go
zreigz: done

internal/controller/stackrunjob_controller.go
zreigz: done

internal/controller/stackrunjob_controller.go
zreigz: done

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):

PR #203, opened 2024-05-29
feat(harness): add support for tf plural managed state

floreks:
Fixes ENG-2126 ENG-2125

### DIFF
--- a/pkg/harness/stackrun/v1/types.go
+++ b/pkg/harness/stackrun/v1/types.go
@@ -17,6 +17,9 @@ type StackRun struct {
 	ExecWorkDir *string
 	Approval    bool
 	ApprovedAt  *string
+	ManageState bool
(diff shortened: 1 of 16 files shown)


pkg/harness/stackrun/v1/types.go
```diff
@@ -17,6 +17,9 @@ type StackRun struct {
 	ExecWorkDir *string
 	Approval    bool
 	ApprovedAt  *string
+	ManageState bool
+	User       *gqlclient.UserFragment
```
michaeljguarino, line 21: there's a different field you should sideload for plural provider auth, `pluralCreds`.  The actual actor user is likely not helpful (beyond ultimately being the user those creds are generated from).

It's also expected this will be nil pretty often in which case set no env vars

michaeljguarino (approved):

PR #207, opened 2024-06-04
feat: Kevin  build Ansible harness image

seemywingz:
Add GHA and DockerFile to build Harnes Ansible image

The final image is based on Python Slim with with Ansible _pip_ installed and the Harness binary copied in


seemywingz:
the [Go Lint Errors](https://github.com/pluralsh/deployment-operator/actions/runs/9373131486/job/25806166134?pr=207) are not related to this PR

.github/workflows/publish-harness.yaml
```diff
@@ -173,3 +173,76 @@ jobs:
...
             HARNESS_BASE_IMAGE_TAG=${{ needs.publish-harness-base.outputs.version }}
             
+  publish-harness-ansible:
+    name: Build and push harness ansible container
+    runs-on: ubuntu-20.04
+    needs: [publish-harness-base]
+    strategy:
+      matrix:
+        versions:
+          - ansible: '4.10.0'
```
michaeljguarino, line 183: For terraform I primarily used major and minor (w/o patch) versions to simplify the user facing config, we should keep that pattern for ansible too. You'd just add another key like `tag` and plumb that in to the docker-metadata action instead of the full ansible version

PR #208, opened 2024-06-05
feat(stack): use default security context for pods/containers

floreks:
- Fixes [PROD-2213](https://linear.app/pluralsh/issue/PROD-2213/stack-controller-seems-to-still-be-setting-environment-variables-to)


pkg/controller/stacks/job.go
```diff
@@ -254,6 +257,28 @@ func ensureDefaultVolume(volumes []corev1.Volume) []corev1.Volume {
...
+}
+
+func ensureDefaultContainerSecurityContext(sc *corev1.SecurityContext) *corev1.SecurityContext {
+	if sc != nil {
+		return sc
+	}
+
+	return &corev1.SecurityContext{
+		AllowPrivilegeEscalation: lo.ToPtr(false),
+		ReadOnlyRootFilesystem:   lo.ToPtr(true),
```
michaeljguarino, line 277: forgot to mention this in the ticket, but this is going to require configuring an emptyDir volume (which we should do of course)

floreks: We've already had one for main `plural` working dir, but I had to add another one for `tmp` dir too.

michaeljguarino (approved):

pkg/controller/stacks/job.go
michaeljguarino: nice

PR #211, opened 2024-06-06
chore: add emptyDir mounted on /tmp for operator and agentk

seemywingz:


charts/deployment-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: deployment-operator
 description: creates a new instance of the plural deployment operator
 appVersion: 0.4.32
-version: 0.4.32
+version: 0.4.33
```
michaeljguarino, line 5: we want to keep appVersion == version here, the chart will be released in the next operator release

seemywingz: okay, I noticed that, but wasn't sure because there were no code changes

michaeljguarino: You can't bump the app version because we haven't published the release.  Just leave this as is.

michaeljguarino: yeah that's not critical at the moment

michaeljguarino (approved):

PR #217, opened 2024-06-12
enhancement: use opentofu binary in harness terraform image

seemywingz:
To avoid violating Terraform's BSL license, we are using the [OpenTofu](https://github.com/opentofu/opentofu) binary

### DIFF
--- a/.github/workflows/publish-harness.yaml
+++ b/.github/workflows/publish-harness.yaml
@@ -96,30 +96,14 @@ jobs:
     runs-on: ubuntu-20.04
     needs: [publish-harness-base]
     env:
-      TERRAFORM_VERSION: 1.8.2
+      TOFU_VERESION: 1.7.2
     strategy:
       matrix:
         versions:
-        - full: 1.8.2
-          tag: 1.8.2
-        - full: 1.8.2
-          tag: "1.8"
-        - full: 1.7.5
-          tag: '1.7'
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-harness.yaml
```diff
@@ -96,29 +96,27 @@ jobs:
...
+        - tofu: 1.6.2
           tag: '1.3'
-        - full: 1.2.9
+        - tofu: 1.6.2
           tag: '1.2'
-        - full: 1.1.9
+        - tofu: 1.6.2
           tag: '1.1'
-        - full: 1.0.11
+        - tofu: 1.6.2
```
michaeljguarino, line 119: are these versions actually aligned?  Looks like the tofu version is always `1.6.2`?

seemywingz: yes, i double checked the [OpenTofu Migration Docs](https://opentofu.org/docs/intro/migration/terraform-1.5-or-lower/)
> When migrating from Terraform 1.5.x or lower, please migrate to OpenTofu 1.6.2 first

michaeljguarino:
Have you tested any runs w/ the tofu versions yet?

seemywingz:
I tested manually. by that I mean downloading the image, volume mounting the `.tf` files and running `terraform plan`

PR #220, opened 2024-06-13
feat: add argo rollouts resume support

zreigz:
- add argo rollouts resume support
 - fix health status for some CRD objects


zreigz:
@michaeljguarino Generally it's ready for review but requires more end-to-end tests. I found the cause of the health status issue and will fix it in this PR. 
I used those manifests for the test: https://github.com/zreigz/argo-rollout-test/tree/main/manifests

internal/controller/argorollout_controller.go
```diff
@@ -0,0 +1,143 @@
...
+		promoteResponse, err := r.CachedClient.Get(promoteURL)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		rollbackResponse, err := r.CachedClient.Get(rollbackURL)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if promoteResponse != closed {
```
michaeljguarino, line 94: Can't we use whether its a 200 response instead of looking at the string response to determine this?

internal/controller/argorollout_controller.go
```diff
@@ -0,0 +1,143 @@
...
+
+		rolloutIf := r.ArgoClientSet.ArgoprojV1alpha1().Rollouts(rollout.Namespace)
+		promoteURL := fmt.Sprintf("%s/ext/v1/gate/%s", consoleURL, serviceID)
+		rollbackURL := fmt.Sprintf("%s/ext/v1/rollback/%s", consoleURL, serviceID)
+
+		promoteResponse, err := r.CachedClient.Get(promoteURL)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		rollbackResponse, err := r.CachedClient.Get(rollbackURL)
```
michaeljguarino, line 89: Caching here might not be best because it'll reduce the responsiveness of the system (might not matter at the end of the day but still should keep it in mind).

internal/controller/argorollout_controller.go
```diff
@@ -0,0 +1,143 @@
...
+			return ctrl.Result{}, err
+		}
+
+		if promoteResponse != closed {
+			return ctrl.Result{}, r.promote(ctx, rolloutIf, rollout)
+
+		}
+		if rollbackResponse != closed {
+			return ctrl.Result{}, r.rollback(rolloutIf, rollout)
+		}
```
michaeljguarino, line 100: I think if the rollout is paused but should neither promote nor rollback, we probably need to requeue it and wait for the user to push either button.

internal/controller/argorollout_controller.go
zreigz: the payload is tiny so I wanted something usable for the future. We could move this cached HTTP client to polly library

internal/controller/argorollout_controller.go
zreigz: it's done here: https://github.com/pluralsh/deployment-operator/pull/220/files#diff-07dca3ddf5eabec847977b5cb0c50ed9b7e51891f146350f2ec630eb6110d7e0R77 
It waits until someone clicks the button

zreigz:
@michaeljguarino PTAL

zreigz:
> I think we should remove the cache and directly use the HTTP status codes as the way to determine if the rollout should be promoted or rolled back. Just feel like that's more robust (and how i designed the api originally).

done

michaeljguarino (approved):

michaeljguarino:
I'm gonna test it a bit but lgtm

michaeljguarino:
Everything seems to have worked, one thing I think might be a nice add would be to kick the service reconciler for the relevant service once you've decided to promote/rollback the rollout though.  Think you should be able to do that pretty simply since you already have the service id.

michaeljguarino:
Ok w/ the svc queue add, it's now really responsive, going to merge it in

PR #222, opened 2024-06-14
fix: do not allow discovery api call for helm templating to stop the agent

floreks:
- Use structured logging for top-level agent error logs
- Use non-deprecated polling function for `APIVersions` map updates
- Bump default number of concurrent reconciles to 20

### DIFF
--- a/pkg/controller/service/reconciler.go
+++ b/pkg/controller/service/reconciler.go
@@ -7,17 +7,6 @@ import (
 	"time"
 
 	console "github.com/pluralsh/console-client-go"
-	clienterrors "github.com/pluralsh/deployment-operator/internal/errors"
-	"github.com/pluralsh/deployment-operator/internal/utils"
-	"github.com/pluralsh/deployment-operator/pkg/applier"
-	"github.com/pluralsh/deployment-operator/pkg/client"
-	"github.com/pluralsh/deployment-operator/pkg/controller"
-	plrlerrors "github.com/pluralsh/deployment-operator/pkg/errors"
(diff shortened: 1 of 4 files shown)


zreigz (approved):

pkg/controller/service/reconciler.go
```diff
@@ -106,13 +107,9 @@ func NewServiceReconciler(ctx context.Context, consoleClient client.Client, conf
...
 		return nil, err
 	}
-	if err := CapabilitiesAPIVersions(discoveryClient); err != nil {
-		return nil, err
-	}
 
 	go func() {
-		//nolint:all
-		_ = wait.PollImmediateInfinite(time.Minute*5, func() (done bool, err error) {
+		_ = wait.PollUntilContextCancel(ctx, time.Minute*5, true, func(_ context.Context) (done bool, err error) {
```
michaeljguarino, line 112: Does this start the polling immediately?  Just want to make sure there isn't a race.  Might even be worth callling `CapabilitiesApiVersions` immediately and swallowing errors to prevent that.

floreks: It does start immediately, there is `true` arg passed that controls that. Anyway, since it is started in a separate routine, the code will continue without waiting. I can adjust that if needed and force calling function to wait.

michaeljguarino: Yeah the only race I can foresee then is the goroutine happens to start after service reconcilers start, so running it synchronously first and discarding errors will at least fix that ordering problem

michaeljguarino (approved):

PR #228, opened 2024-06-26
fix: correctly traverse state module tree

floreks:
- use hashicorp tfjson library to parse tf state 
- correctly traverse state module tree when parsing state resources


pkg/harness/tool/terraform/api/state.go
```diff
@@ -0,0 +1,57 @@
...
+	if resource == nil {
+		return nil
+	}
+
+	return &console.StackStateResourceAttributes{
+		Identifier:    resource.Address,
+		Resource:      resource.Type,
+		Name:          resource.Name,
+		Configuration: lo.ToPtr(ResourceConfiguration(resource)),
+		Links:         algorithms.Map(ResourceLinks(resource), func(d string) *string { return &d }),
```
michaeljguarino, line 55: could use `lo.ToSlicePtr(resource.DependsOn)` for this I think: https://pkg.go.dev/github.com/samber/lo#ToSlicePtr

michaeljguarino (approved):

PR #230, opened 2024-06-28
enhancement: Prod 2299/coreos pod monitor for deployment operator chart

seemywingz:
Update deployment for and enable configuration of deploy-operator Prometheus monitoring


michaeljguarino (commented):
Don't you also need to supply a pod monitor?

PR #235, opened 2024-07-15
feat: add support for enforceNamespace

zreigz:


pkg/controller/service/reconciler.go
```diff
@@ -372,6 +398,8 @@ func (s *ServiceReconciler) Reconcile(ctx context.Context, id string) (result re
 		return
 	}
 
+	manifests = enforceNamespace(manifests, svc.Namespace, svc.SyncConfig)
```
michaeljguarino, line 401: Isn't this actually built into cli utils?  I don't think we need to reinvent this logic.

The `cli-utils` implementation also throws an error which is probably the correct approach instead of silently filtering

zreigz: the `cli-utils` method `manifestreader.SetNamespaces` verifies that every namespaced resource has the namespace, and if one does not, it will set the namespace. The service namespace will throw the error if it already has a different one. I can't override the existing resource namespace with the service namespace

michaeljguarino: Ok I see.  I also think it's likely we want to block creation of cluster-scoped resources entirely if this setting is enabled, which currently would be allowed

zreigz: OK, I have changed it, PTAL

pkg/controller/service/reconciler.go
```diff
@@ -188,6 +188,34 @@ func newDestroyer(invFactory inventory.ClientFactory, f util.Factory) (*apply.De
...
+		return nil
+	}
+
+	for _, man := range mans {
+		gvk := schema.GroupVersionKind{
+			Group:   man.GroupVersionKind().Group,
+			Version: man.GroupVersionKind().Version,
+			Kind:    man.GetKind(),
+		}
+		if controller.ClusterResources.Has(gvk) {
```
michaeljguarino, line 211: what about cluster scoped CRD resources not in the fixed list?

zreigz: I have to add them, thanks

michaeljguarino: Yeah I feel like the best approach here is to use `RESTMapper` which should also give you the ability to determine if the GVK is cluster scoped (or something similar in the k8s discovery client, can't fully remember).

zreigz: yes, I found the way how to do this, PTAL

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):

PR #241, opened 2024-07-19
E2e test

zreigz:
test resource cache

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -1,0 +1,74 @@
+package cache
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


pkg/cache/cache_test.go
```diff
@@ -0,0 +1,70 @@
...
+			Expect(rce.SetSHA(unstructuredPod, ApplySHA)).ToNot(HaveOccurred())
+			Expect(rce.SetSHA(unstructuredPod, ManifestSHA)).ToNot(HaveOccurred())
+			Expect(rce.SetSHA(unstructuredPod, ServerSHA)).ToNot(HaveOccurred())
+
+			cache.Set(key, rce)
+			cachedResource, ok := cache.Get(key)
+			Expect(ok).To(BeTrue())
+			Expect(cachedResource).To(Equal(rce))
+			// should expire and clean applySHA and manifestSHA
+			time.Sleep(1 * time.Second)
```
maciaszczykm, line 60: Is there a way to get rid of these sleep calls? Looks like potential race condition and it will make tests run longer.

maciaszczykm (approved):
Left one comment related to usage of `Sleep()`.

PR #246, opened 2024-07-26
feat: pass reconciled revision id in service update components

zreigz:


pkg/client/service.go
```diff
@@ -27,8 +27,8 @@ func (c *client) GetService(id string) (*console.GetServiceDeploymentForAgent_Se
 	return resp.ServiceDeployment, nil
 }
 
-func (c *client) UpdateComponents(id string, components []*console.ComponentAttributes, errs []*console.ServiceErrorAttributes) error {
-	_, err := c.consoleClient.UpdateServiceComponents(c.ctx, id, components, errs)
+func (c *client) UpdateComponents(id, revisionID, sha string, components []*console.ComponentAttributes, errs []*console.ServiceErrorAttributes) error {
```
michaeljguarino, line 30: I think sha should actually be nullable here (that's expected for freshly created services

zreigz: done

michaeljguarino (approved):

PR #253, opened 2024-08-22
fix: Handle nil variables properly

michaeljguarino:
Should write some tests against this but for now should fix


floreks (approved):

floreks:
Is tf not working when variables file is empty? It must not consider it a valid JSON.

PR #257, opened 2024-08-27
feat: check pending job's Pod

zreigz:
Check if the job's Pod has started successfully. Otherwise, kill the job and set the `failed` status

### DIFF
--- a/internal/controller/stackrunjob_controller.go
+++ b/internal/controller/stackrunjob_controller.go
@@ -41,6 +41,7 @@ import (
 
 const jobSelector = "stackrun.deployments.plural.sh"
 const jobTimout = time.Minute * 40
+const podTimout = time.Minute * 2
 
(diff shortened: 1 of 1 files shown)


internal/controller/stackrunjob_controller.go
```diff
@@ -41,6 +41,7 @@ import (
 
 const jobSelector = "stackrun.deployments.plural.sh"
 const jobTimout = time.Minute * 40
+const podTimout = time.Minute * 2
```
floreks, line 44: nit: typo
s/podTimout/podTimeout

Same for `jobTimout`

floreks: Should it be extended to i.e. 5-10 minutes? @michaeljguarino

michaeljguarino (approved):

PR #258, opened 2024-08-28
feat: kustomize services with liquid templating

zreigz:
- kustomize services with liquid templating
 - unit test


zreigz:
@michaeljguarino PTAL

pkg/manifests/template/kustomize.go
```diff
@@ -38,8 +43,41 @@ func (k *kustomize) Render(svc *console.GetServiceDeploymentForAgent_ServiceDepl
...
+
+		data, err := os.ReadFile(path)
+		if err != nil {
+			return err
+		}
+		data, err = renderLiquid(data, svc)
+		if err != nil {
+			return fmt.Errorf("templating error in %s: %w", rpath, err)
+		}
+		if err := writeFile(path, data); err != nil {
```
michaeljguarino, line 71: will customize choke if this file is not overwritten?  Feels like we should just `writeFile(newPath, data)` instead of overwriting the original liquid file.  This will also cause it to bork on runs that cache the tarball in the agent.

zreigz: done

pkg/manifests/template/kustomize.go
```diff
@@ -68,11 +68,11 @@ func (k *kustomize) Render(svc *console.GetServiceDeploymentForAgent_ServiceDepl
 		if err != nil {
 			return fmt.Errorf("templating error in %s: %w", rpath, err)
 		}
-		if err := writeFile(path, data); err != nil {
+		newPath := strings.TrimSuffix(path, ".liquid")
+		if err := writeFile(newPath, data); err != nil {
```
michaeljguarino, line 72: mega nit but can't this just be `return writeFile(newPath, data)`?

zreigz: done

michaeljguarino (approved):

PR #259, opened 2024-08-28
feat(agent): add vcluster crd and controller

floreks:
- Added a new virtual cluster CRD and controller
  - Register a new virtual cluster in the API
  - Install the virtual cluster with helm using vcluster chart
  - Install agent inside the created virtual cluster
- Minor changes to the helm logic `template` package to be able to reuse added repositories

### CRD
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: vcluster
---
apiVersion: v1
kind: Secret
metadata:
  name: sebastian-vcluster-basic-console-token
  namespace: vcluster
stringData:
  token: "CHANGEME"
---
kind: VirtualCluster
metadata:
  name: sebastian
  namespace: vcluster
spec:
  kubeconfigRef:
    name: vc-sebastian-vcluster-basic
  credentialsRef:
    name: sebastian-vcluster-basic-console-token
    key: token
```

### CRD (for testing locally)
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: vcluster
---
apiVersion: v1
kind: Secret
metadata:
  name: sebastian-vcluster-basic-console-token
  namespace: vcluster
stringData:
  token: "CHANGEME"
---
kind: VirtualCluster
metadata:
  name: sebastian-vcluster-basic
  namespace: vcluster
spec:
  kubeconfigRef:
    name: vc-sebastian-vcluster-basic
  credentialsRef:
    name: sebastian-vcluster-basic-console-token
    key: token
  helm:
    vcluster:
      values:
        exportKubeConfig:
          server: https://localhost:8443
```


api/v1alpha1/virtualcluster_types.go
```diff
@@ -0,0 +1,262 @@
...
+
+func (in *AgentHelmConfiguration) GetRepoUrl() string {
+	if in == nil || in.RepoUrl == nil {
+		return AgentDefaultRepository
+	}
+
+	return *in.RepoUrl
+}
+
+type VClusterHelmConfiguration struct {
```
michaeljguarino, line 212: Is there a reason why you inline the helm configuration struct?  Suppose it's nice for the types to self-document

floreks: That's just not to add an additional level to access the structure. Initially I have used HelmConfiguration directly, but I wanted to have some methods that can default values required to initialize cluster/agent such as repo url or chart name. This way structure access looks the same from the outside but there are different method implementations.

michaeljguarino:
I think we should still want to support separately created vclusters.  So if we could add a bool field to the spec specifying maybe `external` and bypass the creation of the vcluster helm chart (so either people can manage that themselves, via Plural service deployment or loft.sh's tooling) they would be able to.

floreks:
Ok, so when external flag is set then we only upsert to the API, install the agent using provided kubeconfig ref and skip the vcluster chart installation?

michaeljguarino:
> Ok, so when external flag is set then we only upsert to the API, install the agent using provided kubeconfig ref and skip the vcluster chart installation?

Yeah precisely

michaeljguarino (approved):

PR #268, opened 2024-09-16
feat: Ingress replica CRD

zreigz:


api/v1alpha1/ingressreplica_types.go
```diff
@@ -0,0 +1,52 @@
...
+	// Spec of the IngressReplica
+	// +kubebuilder:validation:Required
+	Spec IngressReplicaSpec `json:"spec"`
+
+	// Status of the IngressReplica
+	// +kubebuilder:validation:Optional
+	Status Status `json:"status,omitempty"`
+}
+
+type IngressReplicaSpec struct {
```
michaeljguarino, line 39: We should also support overriding ingress class here

internal/controller/ingressreplica_controller.go
```diff
@@ -0,0 +1,142 @@
...
+		Spec: networkv1.IngressSpec{
+			IngressClassName: oldIngress.Spec.IngressClassName,
+			DefaultBackend:   oldIngress.Spec.DefaultBackend,
+		},
+	}
+	updateIngress(ingressReplica, newIngress, oldIngress)
+	return newIngress
+}
+
+func updateIngress(ingressReplica *v1alpha1.IngressReplica, newIngress *networkv1.Ingress, oldIngress *networkv1.Ingress) {
```
michaeljguarino, line 127: I think we should have some mechanism to merge in new annotations.  I'm on the fence as to whether we should simply take any annotation from the ingress replica itself and override the annotations on the old ingress or have an attribute on the spec.  Sort of leaning towards the first option.

api/v1alpha1/ingressreplica_types.go
zreigz: done

internal/controller/ingressreplica_controller.go
zreigz: I have used `lo.Assign` method which merges multiple maps from left to right.
```
newIngress.Annotations = lo.Assign(newIngress.Annotations, oldIngress.Annotations, ingressReplica.Annotations)
```

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):

PR #269, opened 2024-09-17
feat: metrics aggregate

zreigz:
- Implement a controller to reconcile a metrics aggregate.
 - Create DebounceReconciler as a wrapper that fulfills the controller-runtime interface. It can be use for every controller to delay requests.


api/v1alpha1/metricsaggregate_types.go
```diff
@@ -0,0 +1,55 @@
...
+// +kubebuilder:subresource:status
+type MetricsAggregate struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec of the MetricsAggregate
+	// +kubebuilder:validation:Required
+	Spec MetricsAggregateSpec `json:"spec"`
+
+	// Status of the IngressReplica
```
michaeljguarino, line 32: comment for IngressReplica here

api/v1alpha1/metricsaggregate_types.go
```diff
@@ -0,0 +1,55 @@
...
+	// +kubebuilder:validation:Optional
+	Status Status `json:"status,omitempty"`
+}
+
+type MetricsAggregateSpec struct {
+	Nodes int `json:"nodes"`
```
michaeljguarino, line 38: are these fields optional?  Feels like this should probably be the status struct tbh

internal/controller/metricsaggregate.go
```diff
@@ -0,0 +1,185 @@
...
+		metrics.Spec.MemoryAvailableBytes += nm.MemoryAvailableBytes
+		metrics.Spec.MemoryTotalBytes += nm.MemoryTotalBytes
+	}
+
+	fraction := float64(metrics.Spec.CPUTotalMillicores) / float64(metrics.Spec.CPUAvailableMillicores) * 100
+	metrics.Spec.CPUUsedPercentage = int64(fraction)
+	fraction = float64(metrics.Spec.MemoryTotalBytes) / float64(metrics.Spec.MemoryAvailableBytes) * 100
+	metrics.Spec.MemoryUsedPercentage = int64(fraction)
+
+	return ctrl.Result{}, reterr
```
michaeljguarino, line 118: shouldn't you be requeueing to update on an interval

michaeljguarino (commented):
minor comments but makes sense overall

michaeljguarino:
Another thing we should do is have the agent insert a default metrics aggregate resource on boot (if it doesn't exist).  I'm not sure we'll be able to do it in the helm chart smoothly, because of annoying CRD application/validation stuff, but it would be nice to have zero-config guarantees its there so the console server can query it.

Can just name the default one as `global` or something.

zreigz:
> Another thing we should do is have the agent insert a default metrics aggregate resource on boot (if it doesn't exist). I'm not sure we'll be able to do it in the helm chart smoothly, because of annoying CRD application/validation stuff, but it would be nice to have zero-config guarantees its there so the console server can query it.
> 
> Can just name the default one as `global` or something.

The `DebounceReconciler` sends empty requests when there is no object. I have added `initGlobalMetricsAggregate` method in the `Reconcile` to discover it and create the global MetricsAggregate object

api/v1alpha1/metricsaggregate_types.go
zreigz: done

api/v1alpha1/metricsaggregate_types.go
zreigz: I've moved to status

internal/controller/metricsaggregate.go
zreigz: done

michaeljguarino (commented):
lgtm, why is the lint action failing though?

zreigz:
Expects to bump the chart version

zreigz:
> chart

it's a chart linter that expects to update the version, we do this only for releases

zreigz:
@michaeljguarino can we merge this?

michaeljguarino (approved):

PR #270, opened 2024-09-17
feat: add console manager supervisor logic w/ restart option

floreks:
- Added supervised controller start functionality that can restart the controller if a last poll/reconcile time indicates that it might have died
- Updated Reconciler interface and Controller implementation to allow it to be restarted
- Removed 3 different logger implementations usage across the codebase and replaced it with a single klog logger.
- poll/refresh/jitter interval args are now correctly used by the controllers
- Cleaned up console Reconcilers. Some struct fields were not used anywhere
- Refactored gate cache queue into a standalone cache that can be safely reused by multiple goroutines now
- Refactored queue usage across console reconcilers to use getter instead of reference to a variable
- Refactored `PollUntilContextCancel` usage in the console controller manager not to rely on our internal method implementation when deciding when to stop polling. Internal method will only return error now that can be logged but the poll function will always return `false, nil` (never stop).
- Added controller restart metric counter to be able to track the number of per controller restarts (if any)


seemywingz (approved):

michaeljguarino (commented):
This mostly lgtm for the most part but I think @zreigz should give it a close review as well.

One thing i'm wondering about though is if the heartbeat approach is unnecessary.  You only really need heartbeats for multi-process communication (monitoring liveness in a distributed system).  Couldn't we have some wrapper class like:

```
type MonitoredRoutine struct {
  alive bool
  Runnable func()
}

func (r *MonitoredRoutine) Run() {
  defer func() { 
    // catch panics
    r.alive = false
  }()
  
  r.Runnable()
}

func (r *MonitoredRoutine) Alive() bool { return r.alive }
```

and then if any of them die, just force restart the controller according to some mechanism?

pkg/controller/controller_manager.go
```diff
@@ -75,55 +75,154 @@ func (cm *ControllerManager) GetReconciler(name string) Reconciler {
...
+	wg.Wait()
+	klog.InfoS("All controllers finished", "name", "console-manager")
+}
+
+func (cm *Manager) startControllerSupervised(ctx context.Context, ctrl *Controller) {
+	internalCtx, cancel := context.WithCancel(ctx)
+	wg := &sync.WaitGroup{}
+
+	// Recheck the controller heartbeat every 30 seconds.
+	heartbeatCheckInterval := 30 * time.Second
```
maciaszczykm, line 136: We should probably have argument for that as well. With lower poll interval and jitter it could cause issues. Perhaps adding a check if it is not too short also makes sense.

pkg/controller/controller_manager.go
```diff
@@ -75,55 +75,154 @@ func (cm *ControllerManager) GetReconciler(name string) Reconciler {
...
+	// Make last heartbeat deadline 3 times the time of regular poll.
+	// It means that the controller poller skipped at least the last 2 scheduled polls.
+	// It could indicate that the controller poll might have died and controller should be restarted.
+	lastHeartbeatDeadline := 3 * (cm.PollInterval + cm.Jitter)
```
maciaszczykm, line 140: I feel like there should be an option to change that. Perhaps we should add argument for it and check if it is not too short during initialisation, or have an argument for number of scheduled polls. I am also wondering if some issues would not appear if someone would use short interval and jitter as API requests can take a while.

maciaszczykm: It's not very realistic but still I'd add option to change that later.

maciaszczykm:
I reviewed as well, then we talked about it with @floreks and @zreigz. It looks good to me, issues with pollers being stuck for any reasons should not happen anymore. One thing that can be added is validation for args to avoid situations like poll interval or jitter being too short.

zreigz (approved):

maciaszczykm (approved):

PR #283, opened 2024-09-25
fix: CI github action

floreks:
Releasing dev image from main was failing for a long time due to broken semantic release configuration. Got rid of it as we don't need it on main.


michaeljguarino (approved):
Does the standard tag based release still work? (Pretty sure this is fine just curious)

floreks:
oh yeah, this was a part of the CI action. It was supposed to create the release on every push to the main, but we have decided to rely on manual tag creation later on. The publish action in a separate file handles our standard release process.

PR #290, opened 2024-09-30
fix: metricsaggregate sporadic errors

zreigz:
Replace controller runtime client with dedicated metrics client to avoid resource cache


michaeljguarino (approved):
i'm guessing the errors was from controller-runtime's auto-caching?

PR #295, opened 2024-10-04
fix: helm templating

zreigz:
- read from values.yaml
 - extent API versions for `group/version`
 - add unit test to check `.Capabilities.APIVersions.Has "monitoring.coreos.com/v1"` condition


pkg/manifests/template/helm.go
```diff
@@ -168,7 +168,11 @@ func (h *helm) values(svc *console.GetServiceDeploymentForAgent_ServiceDeploymen
 		}
 	}
 
-	overrides, err := h.valuesFile(svc, "values.yaml.static")
+	filename := "values.yaml.static"
```
michaeljguarino, line 171: what's the issue this is trying to protect against?

zreigz: I have `values.yaml` file in my test scenario. I've noticed it was never taken into consideration. Finally my values map was always empty

michaeljguarino: I feel like that's definitely unusual, i'm pretty sure base values.yaml is always there just not in the values overrides

zreigz: I've changed my test `values.yaml` to `values.yaml.static`

seemywingz (approved):

PR #296, opened 2024-10-04
feat(upgradeinsights): add support for warning status

floreks:


maciaszczykm (approved):

michaeljguarino (approved):

michaeljguarino (commented):
we'd also want to ship the client info w/ user agent information now too w/ the later console api that supports that.

PR #300, opened 2024-10-14
feat: resolve imports from services

zreigz:


pkg/manifests/template/utils.go
```diff
@@ -51,3 +51,14 @@ func contexts(svc *console.GetServiceDeploymentForAgent_ServiceDeployment) map[s
 	}
 	return res
 }
+
+func imports(svc *console.GetServiceDeploymentForAgent_ServiceDeployment) map[string]map[string]string {
+	res := map[string]map[string]string{}
+	for _, imp := range svc.Imports {
+		res[imp.Stack.Name] = map[string]string{}
+		for _, out := range imp.Outputs {
+			res[imp.Stack.Name][out.Name] = out.Value
```
michaeljguarino, line 60: output values are always strings currently?  We probably should document that when it's time to.

michaeljguarino (approved):

pkg/manifests/template/utils.go
zreigz: according to the schema the output values are always strings

PR #301, opened 2024-10-16
feat: reconcile helm resources with hooks + set a healthy state as a default for CRD without a condition block

zreigz:
- add `IgnoreHooks` for helm templating
 - set a healthy state as a default for CRD without a condition block
 - add unit tests to cover new functionalities


michaeljguarino (commented):
Are the hooks prioritized in any way at the moment if they were to be ignored?

Also I think the best solve here is to have a field in a services `helm` struct that determines if the agent reconciles this value, I can work on adding that in the gql api

zreigz:
> Are the hooks prioritized in any way at the moment if they were not ignored? Eg will it order pre-install hooks first?

yes, they are prioritized by type and weight

michaeljguarino (approved):

michaeljguarino (approved):

PR #307, opened 2024-10-28
feat: Add AI insight component scraper

zreigz:


pkg/scraper/component_scraper.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+func setUnhealthyComponents(ctx context.Context, k8sClient ctrclient.Client, gvk schema.GroupVersionKind, opts ...ctrclient.ListOption) error {
+	// Create an unstructured list with the desired GVK
+	list := &unstructured.UnstructuredList{}
+	list.SetGroupVersionKind(gvk)
+
+	// List resources
+	if err := k8sClient.List(ctx, list, opts...); err != nil {
```
michaeljguarino, line 72: we should definitely paginate the list here if we can (don't think it should take much of a refactor).  Mostly to reduce a lot of pressure on the k8s apiserver

pkg/ping/build.go
```diff
@@ -3,16 +3,42 @@ package ping
...
 	console "github.com/pluralsh/console/go/client"
 	"github.com/samber/lo"
 	"k8s.io/apimachinery/pkg/version"
 )
 
 func pingAttributes(info *version.Info, pods []string, minKubeletVersion *string) console.ClusterPing {
 	vs := strings.Split(info.GitVersion, "-")
+	var insightComponents []*console.ClusterInsightComponentAttributes
+	if !scraper.AiInsightComponents.IsEmpty() {
+		insightComponents = make([]*console.ClusterInsightComponentAttributes, 0)
```
michaeljguarino, line 17: minor things, if we send an empty list, it'll actually wipe the components on the server, so if empty just don't add to the cluster ping request at all (I'm hoping it has omitempty on it).  Also i think the full insight list should be fresh or not fresh, not individual ones, that way you can control easily whether to add the whole thing to the ping request.

pkg/scraper/component_scraper.go
zreigz: done, I have used our `algorithms.NewPager`for it

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):
overall lgtm, one thing, I added a ticket, we should add node to this list (might need a direct health implementation), but it would be super useful to get health insights for those objects as well to get a heads up on node health issues.

PR #314, opened 2024-11-19
feat: add custom status readers

zreigz:
Get status directly from the Deployment, ReplicaSet and StatefulSet objects

### DIFF
--- a/internal/kstatus/statusreaders/common.go
+++ b/internal/kstatus/statusreaders/common.go
@@ -1,0 +1,143 @@
+package statusreaders
+
+import (
+	"context"
+	"errors"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/engine"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+// baseStatusReader is the implementation of the StatusReader interface defined
+// in the engine package. It contains the basic logic needed for every resource.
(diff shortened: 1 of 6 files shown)


internal/kstatus/statusreaders/common.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (b *baseStatusReader) Supports(gk schema.GroupKind) bool {
+	return b.resourceStatusReader.Supports(gk)
+}
+
+// ReadStatus reads the object identified by the passed-in identifier and computes it's status. It reads
+// the resource here, but computing status is delegated to the ReadStatusForObject function.
+func (b *baseStatusReader) ReadStatus(ctx context.Context, reader engine.ClusterReader, identifier object.ObjMetadata) (*event.ResourceStatus, error) {
+	object, err := b.lookupResource(ctx, reader, identifier)
```
michaeljguarino, line 51: why is this lookup needed?  If it's needed that's ultimately fine, but seems like the unstructured resource should have been available and could have been avoided.

internal/kstatus/statusreaders/deployment.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (d *deploymentResourceReader) Supports(gk schema.GroupKind) bool {
+	return gk == appsv1.SchemeGroupVersion.WithKind("Deployment").GroupKind()
+}
+
+func (d *deploymentResourceReader) ReadStatusForObject(_ context.Context, _ engine.ClusterReader,
+	deployment *unstructured.Unstructured) (*event.ResourceStatus, error) {
+	identifier := object.UnstructuredToObjMetadata(deployment)
+
+	res, err := status.Compute(deployment)
```
michaeljguarino, line 42: isn't the code that lists the pods w/in kstatus?  I would assume this just means we'll execute it by delegating to it.

internal/kstatus/statusreaders/common.go
zreigz: It resolves the GroupKindVersion for the given identifier and check if the resource exists

internal/kstatus/statusreaders/deployment.go
zreigz: This code looks at the properties that are available on all or most of the Kubernetes resources. I got rid of the code for checking the Pod's status. It checks only the main resource type

michaeljguarino (approved):

PR #320, opened 2024-12-17
feat: Add KubecostExtractors controller

zreigz:


internal/controller/kubecostextractor_controller.go
```diff
@@ -0,0 +1,380 @@
...
+	oneMonthBefore := now.AddDate(0, -1, 0) // 0 years, -1 month, 0 days
+	window := fmt.Sprintf("%d,%d", oneMonthBefore.Unix(), now.Unix())
+
+	queryParams := map[string]string{
+		"window":     window,
+		"aggregate":  aggregate,
+		"accumulate": "true",
+	}
+
+	bytes, err := r.KubeClient.CoreV1().Services(srv.Namespace).ProxyGet("", srv.Name, servicePort, "/model/allocation", queryParams).DoRaw(ctx)
```
michaeljguarino, line 186: Why do we need to use a proxy get here?  Shouldn't we just be able to use the service dns itself?

zreigz: Generally, I got this from the `kubectl-cost` https://github.com/kubecost/kubectl-cost/blob/main/pkg/query/allocation.go#L38

Finally `DoRaw` makes the regular HTTP call

michaeljguarino (approved):

PR #324, opened 2024-12-19
feat: get node and control plane costs

zreigz:
- add nodeCost and controlPlaneCost to cluster cost scrapes
- add jitter to kubecost extraction
- add `enable-kubecost-proxy` flag

### DIFF
--- a/internal/controller/kubecostextractor_controller.go
+++ b/internal/controller/kubecostextractor_controller.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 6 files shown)


zreigz:
I've managed to get control plane cost from `controller` with "__unallocated__" name:
image

internal/controller/kubecostextractor_controller.go
```diff
@@ -69,7 +73,7 @@ func (r *KubecostExtractorReconciler) RunOnInterval(ctx context.Context, key str
 	r.Tasks.Set(key, cancel)
 
 	go func() {
-		_ = wait.PollUntilContextCancel(ctxCancel, interval, true, condition)
+		_ = wait.PollUntilContextCancel(ctxCancel, interval+time.Duration(rand.Int63n(int64(kubeCostJitter))), true, condition)
```
michaeljguarino, line 76: I don't think this will fully work because it will keep the jitter constant through the poll loop.  I'd rather us sleep for the jitter amount at the top of the loop, making it random each run.

michaeljguarino (approved):

internal/controller/kubecostextractor_controller.go
zreigz: yes, you are right, I will change it

zreigz: done

michaeljguarino (approved):

PR #350, opened 2025-01-10
chore: bump agentk version in helm chart

floreks:
Bump agentk to the latest version `0.0.2` with fixed CVE's

### DIFF
--- a/charts/deployment-operator/Chart.yaml
+++ b/charts/deployment-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: deployment-operator
 description: creates a new instance of the plural deployment operator
 appVersion: 0.5.6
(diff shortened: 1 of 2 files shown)


charts/deployment-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: deployment-operator
 description: creates a new instance of the plural deployment operator
 appVersion: 0.5.6
-version: 0.5.6
+version: 0.5.7
```
michaeljguarino, line 5: don't bump this quite yet, because we need it to stay equal to app version to mke deployment operator releases make sense

floreks: Ye, I wasn't sure abut that. Reverted.

PR #358, opened 2025-01-29
fix: kubecost extractor

zreigz:
- fetch resource requests directly from the workload object
 - wipe service status if a service is cached and an error occurs


zreigz:
> why are these fips builds failing btw?

it was flaky, I have rerun and it passed

PR #361, opened 2025-02-03
chore: Optimize deployment operator

zreigz:
- list services for the cluster, w/ a page size <= 50
 - place each listed service in the service cache that was in the agent


michaeljguarino (approved):

michaeljguarino:
We should probably do something similar for stack runs too btw

zreigz:
> We should probably do something similar for stack runs too btw

yes, I will do this also for stack runs

PR #363, opened 2025-02-05
feat: add namespace inference to the runtime service fetch process

zreigz:


pkg/client/cluster.go
```diff
@@ -18,15 +17,63 @@ func (c *client) Ping(vsn string) error {
...
-	for name, version := range svcs {
+	var layouts *console.OperationalLayoutAttributes
+	for name, nv := range svcs {
 		inputs = append(inputs, &console.RuntimeServiceAttributes{
 			Name:    name,
-			Version: version,
+			Version: nv.Version,
 		})
+		switch name {
+		case "cert-manager":
```
michaeljguarino, line 51: can we move these strings to constants (mostly for code cleanliness)

pkg/client/cluster.go
```diff
@@ -18,15 +17,63 @@ func (c *client) Ping(vsn string) error {
...
+func initLayouts(layouts *console.OperationalLayoutAttributes) *console.OperationalLayoutAttributes {
+	if layouts == nil {
+		return &console.OperationalLayoutAttributes{
+			Namespaces: &console.ClusterNamespacesAttributes{},
+		}
+	}
+	return layouts
+}
+
+func appendUniqueExternalDNSNamespace(slice []*string, newValue *string) []*string {
```
michaeljguarino, line 29: lets use a set to build this, then convert it to an array when presenting it to the console api

pkg/client/cluster.go
zreigz: done

pkg/client/cluster.go
zreigz: done

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):

PR #372, opened 2025-02-12
feat: surface upgrade insight errors to conditions

floreks:
Small improvement and cleanup of the upgrade insight controller.

### DIFF
--- a/internal/controller/upgradeinsights_controller.go
+++ b/internal/controller/upgradeinsights_controller.go
@@ -11,7 +11,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	k8sClient "sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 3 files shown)


internal/controller/upgradeinsights_controller.go
```diff
@@ -62,16 +53,19 @@ func (in *UpgradeInsightsController) Reconcile(ctx context.Context, req reconcil
...
+	result := in.handleDelete(ui)
 	if result != nil {
-		return *result, err
+		return *result, nil
 	}
 
 	// Sync UpgradeInsights with the Console API
 	err = in.sync(ctx, ui)
 	if err != nil {
-		logger.Error(err, "unable to save upgrade insights")
```
michaeljguarino, line 74: will this break the poll loop since the controller reconcile is returning error?

michaeljguarino (approved):

internal/controller/upgradeinsights_controller.go
floreks: error has the priority over scheduled reconcile, so it will retry with the backoff limit until err is resolved and then get back to scheduled reconcile loop

PR #375, opened 2025-02-20
feat: add digest url param to service tarball fetches

JohnBlackwell:


pkg/manifests/cache.go
```diff
@@ -53,8 +53,14 @@ func (c *ManifestCache) Fetch(utilFactory util.Factory, svc *console.ServiceDepl
 		return nil, fmt.Errorf("could not fetch tarball url for service")
 	}
 
-	log.V(1).Info("fetching tarball", "url", *svc.Tarball)
-	dir, err := fetch(*svc.Tarball, c.token)
+	log.V(1).Info("fetching tarball", "url", *svc.Tarball, "sha", sha)
+
+	tarballURL := *svc.Tarball
```
michaeljguarino, line 58: I'm pretty sure this url already has params in it, it might be safer to parse it explicitly and add the arg from there.  Also we should make sure it's safely url encoded

michaeljguarino (commented):
we'll definitely want to test this explicitly too, the simplest way is to take the pr tag and apply it to a `deploy-operator` service as the `tag` service secret in the plural ui

pkg/manifests/cache.go
```diff
@@ -55,12 +56,20 @@ func (c *ManifestCache) Fetch(utilFactory util.Factory, svc *console.ServiceDepl
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid tarball URL: %w", err)
+	}
+
 	if sha != "" {
-		tarballURL = fmt.Sprintf("%s?digest=%s", tarballURL, sha)
+		// Get existing query params and safely append the digest
+		q := tarballURL.Query()
+		q.Set("digest", sha)
+		tarballURL.RawQuery = q.Encode()
```
michaeljguarino, line 69: do you actually have to set the raw query or would `tarballURL.String()` have encoded it for you anyway?

michaeljguarino (approved):

michaeljguarino:
Be sure to test this e2e before merging as well

pkg/manifests/cache.go
JohnBlackwell: You have to set raw query. I went ahead and tested that out to make sure.

https://go.dev/play/p/-od7Ejg7nJY

pkg/manifests/cache_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuildTarballURL(t *testing.T) {
+	t.Run("valid URL, no SHA", func(t *testing.T) {
+		u, err := buildTarballURL("https://example.com/foo/bar", "")
```
michaeljguarino, line 11: modify this test case where this url already has a query param, eg `https://example.com/foo/bar?serviceId=some-id` (that's actually how this url is sent from the api)

michaeljguarino (commented):
did not realize this was happening before but in this line https://github.com/pluralsh/deployment-operator/pull/375/files#diff-1fbd1a819df1a915806194907eadaa9f90cc31ef4cd87f8cc9baed69a9e99a41R74 we are fetching the entire tarball into an in-memory string then untarring, this should be read from the `resp.Body` directly which is itself an io reader and much more memory efficient.

Must have been a case of over-reusing code that behaves very differently

pkg/manifests/tarball.go
```diff
@@ -18,7 +17,7 @@ var (
 	client = &http.Client{Timeout: 15 * time.Second}
 )
 
-func get(url, token string) (string, error) {
+func getBody(url, token string) (string, error) {
 	req, err := http.NewRequest(http.MethodGet, url, nil)
```
michaeljguarino, line 21: why not call `getReader` here?

pkg/manifests/tarball.go
```diff
@@ -63,14 +93,14 @@ func fetch(url, token string) (string, error) {
 		return "", err
 	}
 
-	resp, err := get(url, token)
+	resp, err := getReader(url, token)
```
michaeljguarino, line 96: this needs to be `defer resp.Close()`'ed

pkg/manifests/tarball.go
JohnBlackwell: because it's returning a string. this is for the sha.

JohnBlackwell: i could rename it to make it more clear - like `getString`

PR #381, opened 2025-03-10
feat: add service mesh discovery to operational layout

floreks:
Added service mesh discovery to our cache package. It will be refreshed on every discovery cache refresh. Service mesh is detected based on the registered resource group names.

### DIFF
--- a/pkg/cache/discovery_cache.go
+++ b/pkg/cache/discovery_cache.go
@@ -3,19 +3,40 @@ package cache
 import (
 	"context"
 	"fmt"
+	"strings"
+	"sync"
 	"time"
 
 	cmap "github.com/orcaman/concurrent-map/v2"
-	"github.com/pluralsh/deployment-operator/internal/helpers"
-	"github.com/pluralsh/deployment-operator/internal/metrics"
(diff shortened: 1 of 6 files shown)


michaeljguarino (approved):

pkg/cache/discovery_cache.go
```diff
@@ -61,8 +98,24 @@ func updateDiscoveryCache(discoveryClient *discovery.DiscoveryClient) error {
 
 			discoveryCache.Set(fmt.Sprintf("%s/%s", gv.String(), resource.Kind), true)
 			discoveryCache.Set(gv.String(), true)
+
+			checkAndUpdateServiceMesh(gv.Group)
```
michaeljguarino, line 102: this logic can technically overwrite the value in ways we don't want, we'd ideally want istio to take priority over linkerd over cilium

floreks: oh yeah, forgot about adding priority list

floreks: Done

michaeljguarino (approved):

PR #392, opened 2025-03-19
fix: fetch pods status for pending Deployment,DaemonSet and StatefulSet

zreigz:
We don't have complete information about those three workloads. They can be in a pending state forever, and we don't know this. The only way is to additionally check the pods' statuses.

### DIFF
--- a/pkg/scraper/component_scraper.go
+++ b/pkg/scraper/component_scraper.go
@@ -6,10 +6,7 @@ import (
 	"sync"
 	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"github.com/cert-manager/cert-manager/pkg/apis/certmanager"
-
 	"github.com/pluralsh/deployment-operator/internal/helpers"
 	"github.com/pluralsh/deployment-operator/pkg/common"
 	agentcommon "github.com/pluralsh/deployment-operator/pkg/common"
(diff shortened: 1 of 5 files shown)


pkg/scraper/component_scraper.go
```diff
@@ -156,6 +154,27 @@ func setUnhealthyComponents(ctx context.Context, k8sClient ctrclient.Client, gvk
 	return nil
 }
 
+func getResourceHealthStatus(ctx context.Context, k8sClient ctrclient.Client, obj *unstructured.Unstructured) (*common.HealthStatus, error) {
```
michaeljguarino, line 157: Ok i have a better handle on what's happening here.  I still think it's better to air on the side of not fetching pods here, and instead keep an in-memory cache of seen deployments/statefulsets/daemonsets that are in-progress to do the timestamp logic.  Basically it would need to:

* add an entry if in progressing state, with the timestamp of when first found
* remove entry if no longer progressing
* mark as failed if it exceeds a threshold (30m seems fine)

This will mostly be ok since our system keeps insight components around for a while anyway on these and we can tune it and removes the risk of major resource utilization on large clusters.

pkg/scraper/component_scraper.go
```diff
@@ -160,18 +161,42 @@ func getResourceHealthStatus(ctx context.Context, k8sClient ctrclient.Client, ob
...
-	if health.Status == common.HealthStatusProgressing {
-		unhealthyPods, err := common.HasUnhealthyPods(ctx, k8sClient, obj)
-		if err != nil {
-			return nil, err
-		}
-		if unhealthyPods {
-			return &common.HealthStatus{
-				Status: common.HealthStatusDegraded,
-			}, nil
+	if obj.GetAnnotations() == nil {
```
michaeljguarino, line 164: can we refactor extracting the last found timestamp into a separate function (and maybe add the timestamp back in a defer block if not found?).  That will make it more clean to read and reusable.

michaeljguarino (approved):
this looks good as a first pass, it is possible other operators muck with that annotation though, if we see that mght just need to implement it in-memory (whihc i still think is fine)

PR #398, opened 2025-04-10
fix: Support `ignoreCRDs` and fix bug in debounce controller

maciaszczykm:


michaeljguarino (approved):

internal/controller/debounce_controller.go
```diff
@@ -61,7 +61,7 @@ func (r *DebounceReconciler) Start(ctx context.Context) {
 				r.lastRequest = time.Now()
 			case <-ticker.C:
 				// Check if enough time has passed since the last request.
-				if time.Since(r.lastRequest) >= r.debounceDuration {
+				if !r.lastRequest.IsZero() && time.Since(r.lastRequest) >= r.debounceDuration {
```
maciaszczykm, line 64: Due to lack of this check it was possible that first requests were queued before manager was fully initialised. Checked it on the broken cluster and the log line pointing to the debounce controller is gone but the other part of the problem is still there (rate limit issue).

michaeljguarino (approved):

PR #408, opened 2025-04-14
fix: when names have underscores

zreigz:
Replace `_` with `-` for the resource name.
Fix inventory file when the resource was applied


michaeljguarino (commented):
wouldn't this potentially cause conflicts with existing resources since you're going to mark them as not in the inventory with this change?

zreigz:
> wouldn't this potentially cause conflicts with existing resources since you're going to mark them as not in the inventory with this change?

Generally, the cli-utils first create/update inventory before applying. The resources won't be created at all.  I change only faulty items in the inventory file to be able to apply them

PR #409, opened 2025-04-14
chore: Split builds between ARM and AMD64

maciaszczykm:
See: https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners

Results:
Zrzut ekranu 2025-04-14 o 15 31 41
Zrzut ekranu 2025-04-14 o 15 20 34
Zrzut ekranu 2025-04-14 o 14 23 54


michaeljguarino (commented):
why do you need to publish in a separate step?

michaeljguarino (approved):

PR #413, opened 2025-04-16
feat: return recommendations for each pod in the parent object

zreigz:
Add multi-aggregation capability to `getAllocation` to retrieve pod-level allocations across Deployments, StatefulSets, and DaemonSets. Use this enhancement to return recommendations


michaeljguarino (commented):
Could you link to the api docs for this?  I want to double check that it's the intended usage

zreigz:
@michaeljguarino https://www.ibm.com/docs/en/kubecost/self-hosted/2.x?topic=apis-allocation-api
The `aggregate` path parameter

The same used in the UI
image

zreigz:
I debugged the results and confirmed they're returning the parent pods.

zreigz:
@michaeljguarino I'm now preparing recommendation attributes for Deployments, StatefulSets, and DaemonSets, using values taken from their pods.

PR #422, opened 2025-04-24
fix: Manifest cache jitter

maciaszczykm:
- [x] Added 10m (configurable via args) expiry jitter to manifest cache.
- [x] Updated service controller logic to update error if it is nil, so errors from prior runs will not persist.
- [x] Increased tarball fetch timeout from 15s to 1m.

### DIFF
--- a/pkg/manifests/cache.go
+++ b/pkg/manifests/cache.go
@@ -2,6 +2,7 @@ package manifests
 
 import (
 	"fmt"
+	"math/rand"
 	"net/url"
 	"os"
 	"time"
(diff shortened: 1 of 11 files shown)


pkg/manifests/cache.go
```diff
@@ -101,8 +104,8 @@ func (c *ManifestCache) Expire(id string) {
 	c.cache.Remove(id)
 }
 
-func (l *cacheLine) live(dur time.Duration) bool {
-	return l.created.After(time.Now().Add(-dur))
+func (l *cacheLine) live(expiry, jitter time.Duration) bool {
+	return l.created.After(time.Now().Add(-expiry).Add(-time.Duration(rand.Int63n(int64(jitter)))))
```
michaeljguarino, line 108: the jitter should be a stable for ech line, we should add it to the go struct, and set it to a random value between 0 and 10m on create (then when it's refreshed, regenerated the jitter).

This will recompute it each time the line is checked for liveness, which isn't ideal.

maciaszczykm: It's done now. PTAL.

PR #426, opened 2025-05-07
fix: error handling

zreigz:


maciaszczykm (approved):

michaeljguarino (commented):
isn't this going to lose the context that it's a sync error when the defer cleans it up?

zreigz:
> isn't this going to lose the context that it's a sync error when the defer cleans it up?

@michaeljguarino  before the fix the `s.UpdateErrorStatus(ctx, id, err)` in defer function always received nil for the err argument. In this case, the error was added and then deleted in the API. Now when manifests are applied correctly the error is nil and `UpdateErrorStatus`  removes it from the API

I tested it with a failing manifest, and again after fixing the manifest. Everything appears to be working correctly now.

michaeljguarino (approved):
Oh wait nvm, it does set the sync status.

This still does imply two calls to update error statuses though which I think is not ideal

michaeljguarino:
@zreigz made a slight adjustment which i think will prevent double status updates

zreigz:
@michaeljguarino works fine, I can merge it

PR #436, opened 2025-05-19
feat: add component children tree with cache

floreks:
```
TBD
```


zreigz (approved):

pkg/cache/db/cache.go
```diff
@@ -0,0 +1,258 @@
...
+// Children retrieves all child components and their descendants (up to 4 levels deep) for a given component UID.
+// It returns a slice of ComponentChildAttributes containing information about each child component.
+//
+// Parameters:
+//   - uid: The unique identifier of the parent component
+//
+// Returns:
+//   - []ComponentChildAttributes: A slice containing the child components and their attributes
+//   - error: An error if the database operation fails or if the connection cannot be established
+func (in *ComponentCache) Children(uid string) (result []client.ComponentChildAttributes, err error) {
```
michaeljguarino, line 74: since we're doing this in memory it probably doesn't matter, but ideally you'd have a way to fetch children for a list of uids and use that while probing each level of the component tree graph

pkg/cache/db/queries.go
```diff
@@ -0,0 +1,66 @@
...
+			'group' = excluded.'group',
+			version = excluded.version,
+			kind = excluded.kind,
+			namespace = excluded.namespace,
+			name = excluded.name,
+			health = excluded.health
+	`
+
+	componentChildren = `
+		WITH RECURSIVE descendants AS (
```
michaeljguarino, line 51: oh nvm that's the correct way to traverse for sure, didn't realize sqlite supports with recursive queries

michaeljguarino (approved):

PR #440, opened 2025-05-20
feat: (terraform) Add support for parallelism and refresh flags

JohnBlackwell:
Add support for configuring parallelism and refresh flags in Terraform operations:
- Add parallelism and refresh fields to StackRun configuration
- Extend PlanArgsModifier and ApplyArgsModifier to support new flags
- Update Terraform tool to pass parallelism and refresh flags to commands
- Propagate configuration from StackRunBaseFragment to execution

This change allows users to control:
- Number of concurrent operations (parallelism)
- Whether to refresh state before planning/applying (refresh)


pkg/harness/stackrun/v1/types.go
```diff
@@ -60,6 +63,8 @@ func (in *StackRun) FromStackRunBaseFragment(fragment *gqlclient.StackRunBaseFra
 		Variables:    fragment.Variables,
 		PolicyEngine: fragment.PolicyEngine,
 		DryRun:       fragment.DryRun,
+		Parallelism:  fragment.Configuration.Terraform.Parallelism,
```
floreks, line 66: Is it certain that the terraform structure will never be nil? What about ansible or custom runs?

pkg/harness/tool/terraform/terraform.go
```diff
@@ -149,7 +149,7 @@ func (in *Terraform) Scan() ([]*console.StackPolicyViolationAttributes, error) {
 func (in *Terraform) plan() (string, error) {
 	output, err := exec.NewExecutable(
 		"terraform",
-		exec.WithArgs([]string{"show", in.planFileName}),
+		exec.WithArgs([]string{"show", in.planFileName, fmt.Sprintf("-parallelism=%d", *in.parallelism), fmt.Sprintf("-refresh=%t", *in.refresh)}),
```
floreks, line 152: I don't see `parallelism` and `refresh` being assigned in the `Terraform` structure anywhere.

It should be checked for nil before appending and dereferencing.

pkg/harness/stackrun/v1/types.go
JohnBlackwell: good point, i'll add a check. i'm not familiar with ansible/custom runs and how they work with terraform but i can check for nil

michaeljguarino, line 66: Configuration.Terraform can be nil and this will segfault.  It's better to just keep that entire struct as a field and nil check internally for safety

pkg/harness/stackrun/v1/types.go
```diff
@@ -72,12 +77,21 @@ func (in *StackRun) Env() []string {
 		env[i] = fmt.Sprintf("%s=%s", e.Name, e.Value)
 	}
 
+	var tfArgs []string
 	if in.Creds != nil {
```
michaeljguarino, line 81: why are you wiping PLURAL_ACCESS_TOKEN?  That'll also crash the process if it needs plural creds

pkg/harness/stackrun/v1/types.go
```diff
@@ -72,12 +77,21 @@ func (in *StackRun) Env() []string {
...
 		env = append(env, fmt.Sprintf("PLURAL_CONSOLE_URL=%s", lo.FromPtr(in.Creds.URL)))
 	}
 
-	return env
+	if in.Parallelism != nil {
+		tfArgs = append(tfArgs, fmt.Sprintf("-parallelism=%d", *in.Parallelism))
+		env = append(env, fmt.Sprintf("PARALLELISM=%d", *in.Parallelism))
+	}
+
+	tfArgs = append(tfArgs, fmt.Sprintf("-refresh=%t", in.Refresh))
```
michaeljguarino, line 91: don't set this always

pkg/harness/stackrun/v1/types.go
```diff
@@ -43,6 +46,13 @@ func (in *StackRun) MaxSeverity() int {
 }
 
 func (in *StackRun) FromStackRunBaseFragment(fragment *gqlclient.StackRunBaseFragment) *StackRun {
+	var parallelism *int64
```
michaeljguarino, line 49: not the end of the world, but this would be more straightforward to do as:

```elixir
stack := &StackRun{...}
if tf := fragment.Configuration.Terraform; tf != nil {
  stack.Parallelism = tf.Parallelism
  refresh = tf.Refresh
}
```

Also is `Configuration` not a pointer?  If it can be then it can still cause a segfault

michaeljguarino (approved):

pkg/harness/stackrun/v1/types.go
```diff
@@ -77,7 +87,16 @@ func (in *StackRun) Env() []string {
 		env = append(env, fmt.Sprintf("PLURAL_CONSOLE_URL=%s", lo.FromPtr(in.Creds.URL)))
 	}
 
-	return env
+	var tfArgs []string
+	if in.Parallelism != nil {
+		tfArgs = append(tfArgs, fmt.Sprintf("-parallelism=%d", *in.Parallelism))
```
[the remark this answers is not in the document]
JohnBlackwell: resolved: https://github.com/pluralsh/deployment-operator/pull/445

PR #447, opened 2025-05-22
feat: Filter sensitive attributes

maciaszczykm:


pkg/harness/tool/terraform/api/state.go
```diff
@@ -24,11 +26,49 @@ func OutputValueString(value interface{}) string {
...
+	c, err := copystructure.Copy(m)
+	if err != nil {
+		return maps.Clone(m) // Return shallow copy if deep copy fails.
+	}
+
+	return c.(map[string]any)
+}
+
+func ExcludeSensitiveValues(values map[string]any, sensitiveValues map[string]any) map[string]any {
+	out := CloneMap(values)
```
michaeljguarino, line 39: do you really need to clone the map on each level of recursion (it will make this a good bit less performant).  I'm guessing the only really risky part of doing it in place is destabilizing the `for k, v :=  range ...` but you could just put the keys in a new list and range over those perhaps?

floreks: It should be possible to start with an empty map and only append non-sensitive values. thus avoiding direct copy.

maciaszczykm: I have changed it to do it in place.

maciaszczykm: @michaeljguarino PTAL

michaeljguarino (approved):

PR #450, opened 2025-05-27
feat: Add node statistics

maciaszczykm:
Zrzut ekranu 2025-05-28 o 12 55 14


pkg/cache/db/queries.go
```diff
@@ -64,5 +74,32 @@ const (
...
+			node = excluded.node,
+			createdAt = excluded.createdAt
+	`
+
+	nodeStatistics = `
+	SELECT node, COUNT(*)
+	FROM Pod
+	WHERE datetime(createdAt) < datetime('now', '-5 minutes')
+	GROUP BY node
+	`
```
maciaszczykm, line 104: Do we want to change the query so it will return nodes that have no pending pods created more than 5 minutes ago as well?

PR #457, opened 2025-06-06
feat: add lua script support

zreigz:
- Add Lua script support
 - Add missing values from the `helm.values`

### DIFF
--- a/pkg/manifests/template/helm.go
+++ b/pkg/manifests/template/helm.go
@@ -186,12 +186,10 @@ func (h *helm) luaValues(svc *console.ServiceDeploymentForAgent) (map[string]int
 	if svc == nil {
 		return nil, fmt.Errorf("no service found")
 	}
-	if svc.Helm == nil {
+	if svc.Helm == nil || svc.Helm.LuaScript == nil {
 		return newValues, nil
 	}
-	if svc.Helm.LuaScript == nil {
-		return newValues, nil
-	}
+
 	p := luautils.NewProcessor(h.dir)
 	defer p.L.Close()
 


pkg/manifests/template/helm.go
```diff
@@ -170,19 +178,75 @@ func (h *helm) Render(svc *console.ServiceDeploymentForAgent, utilFactory util.F
...
+}
+
 func (h *helm) values(svc *console.ServiceDeploymentForAgent) (map[string]interface{}, error) {
 	currentMap, err := h.valuesFile(svc, "values.yaml.liquid")
 	if err != nil {
 		return currentMap, err
 	}
-	if svc.Helm != nil && svc.Helm.ValuesFiles != nil && len(svc.Helm.ValuesFiles) > 0 {
+	if svc.Helm != nil {
 		for _, f := range svc.Helm.ValuesFiles {
```
michaeljguarino, line 235: what happens if ValuesFiles is nil here?

michaeljguarino: Seems like we should still keep that condition even if it's theoretically covered elsewhere, just in case

zreigz: in Go, range safely handles nil for slices

zreigz: there is no need to double-check it: https://go.dev/play/p/d4h5n4pfSn7
and the `svc.Helm.ValuesFiles` is used only in this place

pkg/manifests/template/helm.go
```diff
@@ -170,19 +178,75 @@ func (h *helm) Render(svc *console.ServiceDeploymentForAgent, utilFactory util.F
...
 
+func (h *helm) luaValues(svc *console.ServiceDeploymentForAgent) (map[string]interface{}, error) {
+	// Initialize empty results
+	newValues := make(map[string]interface{})
+	var valuesFiles []string
+
+	if svc == nil {
+		return nil, fmt.Errorf("no service found")
+	}
+	if svc.Helm == nil {
```
michaeljguarino, line 189: can we compress these to just `svc.Helm == nil || svc.Helm.LuaScript == nil`?

pkg/manifests/template/helm.go
```diff
@@ -170,19 +178,75 @@ func (h *helm) Render(svc *console.ServiceDeploymentForAgent, utilFactory util.F
...
+	valuesTable := p.L.NewTable()
+	p.L.SetGlobal("values", valuesTable)
+
+	valuesFilesTable := p.L.NewTable()
+	p.L.SetGlobal("valuesFiles", valuesFilesTable)
+
+	// Execute the Lua script
+	err := p.L.DoString(*svc.Helm.LuaScript)
+	if err != nil {
+		return nil, err
```
michaeljguarino, line 208: what do the lua parse errors look like?  We might need to wrap the error if they're gnarly

zreigz: Lua errors typically occur if the Lua code passed to DoString is malformed or invalid. It's good to throw it here

pkg/manifests/template/helm.go
zreigz: done

pkg/manifests/template/helm.go
zreigz: Gopher-Lua's parse errors are generally quite descriptive on their own — they mimic the standard Lua interpreter’s error format closely. For example the function `f` is not ended with `end`:
```lua
function f()
```
the error will be:
`lua: main.lua:1: 'end' expected near <eof> (exit status 1)`

PR #463, opened 2025-06-13
chore: detect splitbrain

zreigz:
- detect splitbrain
 -  make /.kube writable
 -  add emptyDir volume for /.kube


michaeljguarino:
> > this lgtm, can we do an e2e test on it though just to confirm no regressions before merging?
> 
> yes sure, you mean the CLI e2e for the `plural up`?

Nah just deploy the latest console and the deploy operator tag to the dev cluster you use and make sure services don't error

PR #464, opened 2025-06-13
fix: Fix cluster health score query

maciaszczykm:
It was done on failing pods table before but after merging it with components table where clause wasn't added. This fixes the bug and adds tests to cover the changes.

### DIFF
--- a/pkg/cache/db/cache_test.go
+++ b/pkg/cache/db/cache_test.go
@@ -332,15 +332,15 @@ func TestComponentCache_HealthScore(t *testing.T) {
 		defer db.GetComponentCache().Close()
 
 		uid := testUID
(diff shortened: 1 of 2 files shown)


michaeljguarino (commented):
I think i thought i spec'ed this differently but I can't find the ticket.  Basically certain issues should have much higher deductions since a plain ratio will always be pretty high.  Something like:

* do the plain resource health ratio as first grade
* any failing cert should deduct 10 points per cert
* any failing resource in kube-system should deduct 20 points
* any failing persistent volume should deduct 10 points
* failure of coredns/aws-cni/anything in istio namespace should deduct 50
* failing ingress controller namespace service should deduct 50 (can probably just use a heuristic with `ingress-nginx`, `traefik`, etc)

This should obviously have a min of 0 score, since the deductions could end up sub 0.

PR #468, opened 2025-06-18
fix: Improve resource cache crd detection

michaeljguarino:
* adds mutex
* hook into crd reconciler to find new crds as they come in and start watching

Fixes PROD-3708

### DIFF
--- a/internal/controller/crd_controller.go
+++ b/internal/controller/crd_controller.go
@@ -19,6 +19,8 @@ package controller
 import (
 	"context"
 
+	"github.com/pluralsh/deployment-operator/pkg/cache"
+	"github.com/pluralsh/deployment-operator/pkg/common"
 	"github.com/pluralsh/polly/containers"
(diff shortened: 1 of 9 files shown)


internal/controller/crd_controller.go
```diff
@@ -68,6 +70,12 @@ func (r *CrdRegisterControllerReconciler) Reconcile(ctx context.Context, req ctr
 			continue
 		}
 
+		if crd.Labels != nil {
+			if _, ok := crd.Labels[common.ManagedByLabel]; !ok {
```
floreks, line 74: shouldn't it be `ok` not `!ok` to check if it is managed by the agent?

Also we can probably move it out of the loop since watchers do not care about version and ignore it anyway

michaeljguarino: i think what it should actually do is build the full set in the loop, will add that

JohnBlackwell (approved):

PR #473, opened 2025-06-23
fix: save manifest sha only if the resource was successfully applied

floreks:


zreigz (approved):

pkg/cache/resource_cache.go
```diff
@@ -229,11 +229,24 @@ func SaveResourceSHA(resource *unstructured.Unstructured, shaType SHAType) {
...
+func CommitManifestSHA(resource *unstructured.Unstructured) {
+	if !initialized {
+		klog.V(4).Info("resource cache not initialized")
+		return
+	}
+
+	key := object.UnstructuredToObjMetadata(resource).String()
+	sha, _ := resourceCache.GetCacheEntry(key)
+	sha.CommitManifestSHA()
+	resourceCache.SetCacheEntry(key, sha)
```
michaeljguarino, line 247: do you actually need to set it back? I'm guessing it's just updating the sha in-place and it's already in the cache hash map right

floreks: It might work without an explicit set, since the underlying fields are pointers. The `GetCacheEntry` returns an object so a shallow copy of the cache entry. Pointers should still reference the same memory though.

michaeljguarino: It's probably fine to keep this as-is now to be safe

PR #476, opened 2025-06-25
fix: remove the incorrect error message

zreigz:


maciaszczykm (approved):

michaeljguarino (commented):
but why were these errors so frequent in the first place?

PR #479, opened 2025-07-02
restart a pod when ephemeral storage is unavailable

zreigz:


charts/deployment-operator/values.yaml
```diff
@@ -114,7 +114,11 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
-resources: {}
+resources:
+  requests:
+    ephemeral-storage: "4Gi"
```
michaeljguarino, line 119: we should be setting cpu/memory requests too (that's in another ticket).

michaeljguarino (commented):
I'm still not positive this is the right root cause for what we were seeing.  What's the reasoning here.

charts/deployment-operator/values.yaml
michaeljguarino: (if this is the root cause, I think we should set 10Gi as our limit just to be safe though)

zreigz:
> I'm still not positive this is the right root cause for what we were seeing. What's the reasoning here.

I haven’t been able to reproduce the issue in my environment, and I don’t have access to the environment where it’s happening.

michaeljguarino:
> > I'm still not positive this is the right root cause for what we were seeing. What's the reasoning here.
> 
> I haven’t been able to reproduce the issue in my environment, and I don’t have access to the environment where it’s happening.

I see it fairly consistently when running the deployment operator in a kind cluster (eg like in the boot-aws console)

zreigz:
> > > I'm still not positive this is the right root cause for what we were seeing. What's the reasoning here.
> > 
> > 
> > I haven’t been able to reproduce the issue in my environment, and I don’t have access to the environment where it’s happening.
> 
> I see it fairly consistently when running the deployment operator in a kind cluster (eg like in the boot-aws console)

I also run an operator in kind cluster but on our test console. I will switch to boot-aws console, thanks

PR #488, opened 2025-07-16
wip: add ability to ignore field in a resource by json path

zreigz:
This PR introduces the ignoreUpdateFields mechanism to support drift normalization by ignoring updates on specific subfields of the service's resources. 
 - Resources can define ignored fields via the `deployments.plural.sh/ignore-fields` annotation. The annotation accepts a comma-separated list of JSON paths (e.g. /spec/replicas,/metadata/labels).
 - Also supports ignoring fields via the `ServiceDeployment.Spec.SyncConfig.DiffNormalizers` field, providing more centralized configuration.

### DIFF
--- a/pkg/controller/service/normalizing.go
+++ b/pkg/controller/service/normalizing.go
@@ -1,0 +1,73 @@
+package service
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"github.com/xeipuuv/gojsonpointer"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func IgnoreJSONPaths(ctx context.Context, c client.Client, obj unstructured.Unstructured, ignorePaths []string) (unstructured.Unstructured, error) {
+	current := &unstructured.Unstructured{}
+	current.SetGroupVersionKind(obj.GroupVersionKind())
+
+	key := types.NamespacedName{
+		Name:      obj.GetName(),
+		Namespace: obj.GetNamespace(),
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/controller/service/normalizing.go
```diff
@@ -71,3 +72,68 @@ func matchesKey(obj unstructured.Unstructured, key normalizerKey) bool {
...
+		return float64(n)
+	case uint16:
+		return float64(n)
+	case uint32:
+		return float64(n)
+	case uint64:
+		return float64(n)
+	case float32:
+		return float64(n)
+	case float64:
```
maciaszczykm, line 134: You can use the same case as in `isNumber`:

```go
case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, float32, float64:
```

zreigz: Unfortunately, I can't do this:
image

michaeljguarino:
My recent console pr adds the ability to control whether this should backfill or prune the json pointer.  Maybe lets get that merge and add that logic in this pr directly.

This is the code: https://github.com/pluralsh/console/pull/2463/files#diff-e6f35c4c9fc8b6eb1ce487d317e255f7915d6f9703c85b65152e3ce9e0435f01R62

pkg/controller/service/normalizing.go
```diff
@@ -2,18 +2,58 @@ package service
...
+	if err != nil {
+		return obj, fmt.Errorf("failed to marshal patch: %w", err)
+	}
+
+	raw, err := json.Marshal(obj.Object)
+	if err != nil {
+		return obj, fmt.Errorf("failed to marshal object: %w", err)
+	}
+
+	patch, err := jsonpatch.DecodePatch(patchJSON)
```
michaeljguarino, line 37: is there no way we can manually create this json patch struct w/o routing it through a json representation?

michaeljguarino (approved):
minor changes

PR #490, opened 2025-07-22
feat: openshift distro discovery

zreigz:
I was able to install a local OpenShift cluster and test it

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -13,6 +13,7 @@ import (
 	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	templatesv1 "github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1"
 	constraintstatusv1beta1 "github.com/open-policy-agent/gatekeeper/v3/apis/status/v1beta1"
+	openshift "github.com/openshift/api/config/v1"
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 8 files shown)


cmd/agent/main.go
```diff
@@ -47,6 +48,7 @@ func init() {
 	utilruntime.Must(templatesv1.AddToScheme(scheme))
 	utilruntime.Must(rolloutv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(certmanagerv1.AddToScheme(scheme))
+	utilruntime.Must(openshift.AddToScheme(scheme))
```
michaeljguarino, line 51: This won't fail on non-openshift right?

pkg/common/k8s.go
```diff
@@ -150,3 +151,21 @@ func SupportedMetricsAPIVersionAvailable(discoveredAPIGroups *metav1.APIGroupLis
...
+	return false
+}
+
+func GetOpenShiftVersion(c k8sClient.Client) (string, error) {
+	cv := &configv1.ClusterVersion{}
+	err := c.Get(context.Background(), k8sClient.ObjectKey{Name: "version"}, cv)
+	if err != nil {
+		return "", err
+	}
+	return cv.Status.Desired.Version, nil
```
michaeljguarino, line 170: is there not a current version?  This will mark it as the intended version during upgrade i'm guessing

pkg/ping/build.go
```diff
@@ -32,14 +32,15 @@ func pingAttributes(info *version.Info, pods []string, minKubeletVersion *string
...
-		KubeletVersion: minKubeletVersion,
-		Distro:         lo.ToPtr(findDistro(append(pods, info.GitVersion))),
-		HealthScore:    &hs,
-		NodeCount:      &nodCount,
-		PodCount:       podCount,
-		NamespaceCount: &namespaceCount,
-		NodeStatistics: ns,
+		CurrentVersion:   strings.TrimPrefix(vs[0], "v"),
+		KubeletVersion:   minKubeletVersion,
+		Distro:           lo.ToPtr(findDistro(append(pods, info.GitVersion))),
```
michaeljguarino, line 37: don't believe this is setting distro to OPENSHIFT

zreigz: I missed it.

zreigz: OPENSHIFT is missing on the API site

PR #505, opened 2025-08-06
feat: add AgentConfiguration crd

zreigz:
https://console.plrl-dev-aws.onplural.sh/

AgentConfiguration CRD for deploy-operator


api/v1alpha1/agentconfig_types.go
```diff
@@ -0,0 +1,45 @@
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type AgentConfigurationSpec struct {
```
michaeljguarino, line 8: add servicePollInterval to this

api/v1alpha1/agentconfig_types.go
```diff
@@ -0,0 +1,45 @@
...
+	ClusterPingInterval               *string `json:"clusterPingInterval,omitempty"`
+	RuntimeServicesPingInterval       *string `json:"runtimeServicesPingInterval,omitempty"`
+	StackPollInterval                 *string `json:"stackPollInterval,omitempty"`
+	PipelineGateInterval              *string `json:"pipelineGateInterval,omitempty"`
```
michaeljguarino, line 12: We should potentially have a way to disable these as well

api/v1alpha1/agentconfig_types.go
```diff
@@ -0,0 +1,45 @@
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type AgentConfigurationSpec struct {
+	ClusterPingInterval               *string `json:"clusterPingInterval,omitempty"`
+	RuntimeServicesPingInterval       *string `json:"runtimeServicesPingInterval,omitempty"`
```
michaeljguarino, line 10: Lets name this `compatibilityUploadInterval` (users probably don't know the runtime service terminology)

michaeljguarino (commented):
We need to make sure this structure is fully documented as well

api/v1alpha1/agentconfig_types.go
zreigz: I've added such a possibility when you set the interval to "0s"

zreigz: I had to update the poller to support it. There is also the unit test for it

api/v1alpha1/agentconfig_types.go
zreigz: done

internal/helpers/poll.go
```diff
@@ -2,23 +2,163 @@ package helpers
...
+		// guarantee that we don't invoke condition more than once after context has
+		// been cancelled.
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
+		if !sliding {
+			t.Next()
+		}
+		if ok, err := func() (bool, error) {
```
michaeljguarino, line 91: shouldn't this be `done, err`?

internal/helpers/poll.go
```diff
@@ -2,23 +2,163 @@ package helpers
...
+// true, or the condition returns an error. If sliding is true, the period is computed
+// after condition runs. If it is false then period includes the runtime for condition.
+// If immediate is false the first delay happens before any call to condition, if
+// immediate is true the condition will be invoked before waiting and guarantees that
+// the condition is invoked at least once, regardless of whether the context has been
+// cancelled. The returned error is the error returned by the last condition or the
+// context error if the context was terminated.
+//
+// This is the common loop construct for all polling in the wait package.
+func loopConditionUntilContext(ctx context.Context, t wait.Timer, immediate, sliding bool, condition wait.ConditionWithContextFunc) error {
```
michaeljguarino, line 44: should we get rid of the sliding param?  It looks like it's not used and somewhat confusing

michaeljguarino (commented):
minor comments, think we'll push this into next release just for simplicity anyway

internal/helpers/poll.go
zreigz: done

internal/helpers/poll.go
zreigz: done

michaeljguarino:
This mostly looks good, but I think we should have it bake in a cluster for at least a day before merging, and ideally do a few service creates/deletes through that time to ensure the polls still work as expected

zreigz:
> This mostly looks good, but I think we should have it bake in a cluster for at least a day before merging, and ideally do a few service creates/deletes through that time to ensure the polls still work as expected

I have deployed it on our dev env: `console.plrl-dev-aws.onplural.sh`

internal/controller/agentconfiguration_controller.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		if !field.IsNil() {
+			val := field.Elem().Interface()
+			logger.Info("Overridden config from CRD", fieldName, val)
+		}
+	}
+}
+
+func findComma(tag string) int {
```
maciaszczykm, line 97: You can use [strings.Index](https://pkg.go.dev/strings#Index) or [strings.IndexRune](https://pkg.go.dev/strings#IndexRune).

internal/controller/agentconfiguration_controller.go
```diff
@@ -0,0 +1,104 @@
...
+	logger := log.FromContext(ctx)
+	v := reflect.ValueOf(spec)
+	t := reflect.TypeOf(spec)
+
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i)
+		fieldType := t.Field(i)
+		fieldName := fieldType.Tag.Get("json")
+
+		if commaIdx := len(fieldName); commaIdx > 0 {
```
maciaszczykm, line 84: `commaIdx` seems to be unused. `len(fieldName) > 0` should do the same thing.

internal/controller/agentconfiguration_controller.go
```diff
@@ -0,0 +1,104 @@
...
+// SetupWithManager sets up the controller with the Manager.
+func (r *AgentConfigurationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
+		For(&v1alpha1.AgentConfiguration{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Complete(r)
+}
+
+func LogOverriddenCRDValues(ctx context.Context, spec v1alpha1.AgentConfigurationSpec) {
+	logger := log.FromContext(ctx)
```
maciaszczykm, line 75: Wouldn't logger handle displaying it? It looks like spec already has omitempty annotations set so it will print just the fields that are set (overridden).

internal/helpers/poll.go
floreks, line 1: We can probably just use a wrapper with outer poller that will track interval changes and restart actual poller instead of copying and modifying k8s wait package.

POC
```
package main

import (
	"context"
	"log"
	"time"

	"k8s.io/apimachinery/pkg/util/wait"
)

// DynamicDuration returns a dynamic duration based on the current time.
// It changes every 30 seconds.
func DynamicDuration() time.Duration {
	now := time.Now().Truncate(30 * time.Second).Unix()
	interval := (now / 30) % 30
	return time.Duration(interval+1) * time.Second
}

func DynamicPollUntilContextCancel(ctx context.Context, intervalFunc func() time.Duration, immediate bool, condition wait.ConditionWithContextFunc) error {
	started := false
	lastInterval := intervalFunc()

	return wait.PollUntilContextCancel(ctx, 10 * time.Second, true, func(ctx context.Context) (done bool, err error) {
		log.Println("running global poll")
		interval := intervalFunc()
		if lastInterval == interval && started {
			log.Println("global - interval not changed, returning", "interval", interval)
			return false, nil
		}

		log.Println("global - interval changed, restarting", "lastInterval", lastInterval, "interval", interval)
		started = true
		lastInterval = interval

		go wait.PollUntilContextCancel(ctx, interval, immediate, func(ctx context.Context) (bool, error) {
			log.Println("running local poll")
			curInterval := intervalFunc()
			if interval != curInterval {
				log.Println("local - interval changed, restarting", "curInterval", curInterval, "interval", interval)
				return true, nil
			}

			log.Println("running condition")
			return condition(ctx)
		})
		return false, nil
	})
}

func main() {
	DynamicPollUntilContextCancel(context.Background(), DynamicDuration, true,
		func(ctx context.Context) (done bool, err error) {
			log.Println("executed custom condition")
			return false, nil
		})

	select {}
}
```

zreigz: Good point, thanks

michaeljguarino (approved):
If this has been running consistently for a while, should be good to go

zreigz:
> If this has been running consistently for a while, should be good to go
Yes, it's ready

PR #516, opened 2025-08-18
feat: Replace cli-utils

zreigz:
## Changes
- Added synchronizers that watch for k8s resource changes and sync it to the db store
- Rewritten database store to handle resource cache responsibilities too
  - Resource SHAs to properly skip cached resources during apply
  - Resource relationships to find out dependencies of resources managed by the agent
  - Resource health statuses
  - Service ID for resources managed by the agent 
- Added synchronizers supervisor that can dynamically start/stop/restart synchronizers. It will use a new discovery cache to react to GVK changes in the API and handle synchronizers accordingly.
- Rewritten discovery cache and added a separate manager for it that handles refresh. Cache itself exposes callbacks that can be used to react to changes (add/delete) in the cache.
- Added a new applier with wave support:
  - Ensures that applied resources contain service annotation (used as owning inventory)
  - Handles resource filtering based on custom filters (i.e. skip unchanged cached resources)
  - Splits resource manifests into waves:
    - Wave 0 - core non-namespaced resources (i.e. namespace)
    - Wave 1 - core namespaced configuration resources (i.e. configmap, secret)
    - Wave 2 - core namespaced workload resources (i.e. deployment, job)
    - Wave 3 - core namespaced networking resources (i.e. service, ingress)
    - Wave 4 - all other resources including CRD objects
    - Wave 5 - resources scheduled for deletion
  - Waves are applied in order and resources inside the same wave are all applied concurrently. Max concurrent applies can be controlled to bulk apply resources.
- Got rid of discovery client usage in most places and replaced it with discovery cache to minimize API pressure.
- Added a global `jitter-factor` flag that can control jitter across agent intervals. By default it is set to `0.2` so for a `1s` interval it could return anything between 0.8-1.2s.

## Test Plan

Test environment: https://console.plrl-dev-aws.onplural.sh/

## Checklist

    - [x] Agent logs do not contain any errors.


maciaszczykm (commented):
Checked all files. Sending for review.

floreks (approved):

PR #522, opened 2025-09-01
replace manual timers in poller with apimachinery wait utilities

zreigz:
Every call to `time.After` or `time.NewTimer` allocates a timer and an associated internal goroutine. In multiple threads or loops, these goroutines can accumulate if timers aren’t stopped properly. 
 - The k8s `wait` package avoids this by reusing timers and not leaking goroutines when polling.
 - The k8s `wait` functions integrate `runtime.HandleCrashWithContext`, which ensures that panics in callbacks don’t bring down the whole program.

## Test Plan

Test environment: https://console.your-env.onplural.sh/

## Checklist

- [ ] I have added a meaningful title and summary to convey the impact of this PR to a user.
- [ ] I have deployed the agent to a test environment and verified that it works as expected.
    - [ ] Agent started successfully.
    - [ ] Logs are clean and do not contain errors.
    - [ ] Component trees are working as expected.
- [ ] I have added tests to cover my changes.
- [ ] If required, I have updated the Plural documentation accordingly.

### DIFF
--- a/internal/helpers/poll.go
+++ b/internal/helpers/poll.go
@@ -5,8 +5,6 @@ import (
 	"math/rand"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/runtime"
-
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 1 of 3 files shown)


michaeljguarino:
I'm pretty sure that issue is no longer the case in go 1.23, right?

michaeljguarino (approved):
if this is the right solve that's fine, but i thought the timer code was no longer an issue right?

zreigz:
> if this is the right solve that's fine, but i thought the timer code was no longer an issue right?

Overall, the current solution relies on the same library and logic as before the change. Unit tests cover the functionality used in our code, and manual tests show that everything works, so I can't identify any edge cases that aren't working correctly. This poller is also part of the manager, along with the supervisor. I'll test it as an integrated part to make sure that everything works correctly

PR #529, opened 2025-09-20
Fix crash on condition inference

michaeljguarino:
Wasn't delicate enough in typecasts

## Test Plan
n/a

## Checklist

- [ ] I have added a meaningful title and summary to convey the impact of this PR to a user.
- [ ] I have deployed the agent to a test environment and verified that it works as expected.
    - [ ] Agent starts successfully.
    - [ ] Service creation works without any issues when using raw manifests and Helm templates.
    - [ ] Service creation works when resources contain both CRD and CRD instances.
    - [ ] Service templating works correctly.
    - [ ] Service errors are reported properly and visible in the UI.
    - [ ] Service updates are reflected properly in the cluster.
    - [ ] Service resync triggers immediately and works as expected.
    - [ ] Service deletion works and cleanups resources properly.
    - [ ] Services can be recreated after deletion.
    - [ ] Service detachment works and keeps resources unaffected.
    - [ ] Services can be recreated after detachment.
    - [ ] Service component trees are working as expected.
    - [ ] Cluster health statuses are being updated.
    - [ ] Agent logs do not contain any errors.
- [ ] I have added tests to cover my changes.
- [ ] If required, I have updated the Plural documentation accordingly.


michaeljguarino:
This fixes the given bug but likely also revealed another one, which was that the agent's rbac configurations were wiped when i changed the tag.

michaeljguarino:
The openshift issue is fixed with the mirror groups setup in the applier code, but the solution isn't fully general.  @zreigz et al, what are we doing about subresources created by controllers, eg:

* deployment operator creates a resource with inventory annotation
* operator creates child resources that maintain that annotation for whatever reason
* do we try to delete it on the next service reconcile pass? our cache will likely pick it up as "part" of the service

The obvious solution here is to not cache resources with owner references, just not sure that's viable.

floreks:
Ye, that seems to definitely be another special case that we need to handle. If controller copies annotations from the parent to created objects, it will currently get picked up by the synchronizers and serviceID will be saved in cache entry. We should probably ignore annotation when synchronizing from watches and only save it when directly applying w/ the applier. This is the only way to be sure that resources directly created via incoming manifests are the only ones w/ serviceID in our cache.

michaeljguarino:
> Ye, that seems to definitely be another special case that we need to handle. If controller copies annotations from the parent to created objects, it will currently get picked up by the synchronizers and serviceID will be saved in cache entry. We should probably ignore annotation when synchronizing from watches and only save it when directly applying w/ the applier. This is the only way to be sure that resources directly created via incoming manifests are the only ones w/ serviceID in our cache.
> 
> Right now applier will get all resources w/ serviceID from cache, compare them with manifests and delete extra ones from the cache.

I feel like if we ignore entirely you'll miss the ability to get dependent resources (unless that's a separate table currently).  That's why a boolean column would make sense to me to track applied state.

michaeljguarino:
So thinking about this a bit more:

1. Obviously add the child resource (resource w/ owner references) filtering to the prune logic (we should never prune any child resource, pretty sure this was in the original spec and likely just forgotten).  From what I can tell this is equivalent to argocd too.
2. Consider more delicate approaches, but it's possible (1) is complete

floreks:
Child resources are handled via ownerRef column that stores the UID. serviceID column is only there to compare against incoming manifests so the applier can now what to delete and what to apply.

michaeljguarino:
So we don't add known child resources to the service component table?

zreigz:
We add them to the component children list

PR #530, opened 2025-09-20
fix: Don't warn on crds managed by other services

maciaszczykm:
## Test Plan
Test environment: https://console.plrl-dev-aws.onplural.sh/

## Checklist

    - [x] Agent logs do not contain any errors.

### DIFF
--- a/pkg/streamline/applier/wave.go
+++ b/pkg/streamline/applier/wave.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


michaeljguarino:
> It looks like it does. We probably need to adjust isManaged function instead

I'd just add another nested condition to check if the resource is a crd before adding it to the error channel.  The isManaged function should do what it says (check if the resource isManaged)

floreks (approved):

PR #748, opened 2026-05-05
feat: infracost deployment operator changes

kinjalh:
## Test Plan

https://console.kinjal-gitgud.onplural.sh/

## Checklist

    - [x] Sync waves annotations are respected.
    - [x] Sync phases annotations are respected. Phases are executed in the correct order.
    - [x] Sync hook delete policies are respected. Resources are not recreated once they reach the desired state.
    - [x] Agent logs do not contain any errors (after running for at least 30 minutes).
    - [x] There are no visible anomalies in Datadog (after running for at least 30 minutes).


pkg/harness/tool/terraform/infracost.go
```diff
@@ -0,0 +1,202 @@
...
+	"k8s.io/klog/v2"
+
+	harnessexec "github.com/pluralsh/deployment-operator/pkg/harness/exec"
+	"github.com/pluralsh/deployment-operator/pkg/log"
+)
+
+// Infracost implements [v1.Tool] interface.
+// It runs infracost breakdown on the terraform plan and returns cost estimates.
+func (in *Terraform) Infracost() ([]*console.StackInfracostResourceAttributes, error) {
+	if !in.infracostAvailable() {
```
michaeljguarino, line 34: You should only run infracost if the stack specifies it, not just if the binary is present

pkg/harness/tool/terraform/infracost.go
```diff
@@ -0,0 +1,202 @@
...
+
+	resources := in.convertInfracostReport(report)
+	klog.V(log.LogLevelDebug).InfoS("infracost breakdown completed", "resourceCount", len(resources))
+
+	return resources, nil
+}
+
+// infracostAvailable checks if the infracost binary is available in PATH.
+func (in *Terraform) infracostAvailable() bool {
+	_, err := exec.LookPath("infracost")
```
michaeljguarino, line 64: are we adding infracost to the tf dockerfile?

pkg/harness/tool/v1/tool.go
```diff
@@ -41,6 +41,12 @@ func (in *DefaultTool) HasChanges() (bool, error) {
 	return true, nil
 }
 
+// Infracost implements [Tool] interface.
+// The default implementation returns nil (no infracost support).
+func (in *DefaultTool) Infracost() ([]*console.StackInfracostResourceAttributes, error) {
```
michaeljguarino, line 46: how does this compile considering ansible doesn't implement?

pkg/harness/tool/terraform/infracost.go
```diff
@@ -0,0 +1,202 @@
...
+	return err == nil
+}
+
+// runInfracost executes infracost breakdown and returns the parsed report.
+// Infracost does not accept binary terraform plan files, so we first convert
+// the plan to JSON using 'terraform show -json', write it to a temp file,
+// and then pass that to infracost.
+func (in *Terraform) runInfracost() (*InfracostReport, error) {
+	// Convert binary plan to JSON using terraform show
+	jsonPlan, err := harnessexec.NewExecutable(
```
michaeljguarino: I suspect there's a clever golang way to get this plan into the file w/o it being buffered to an intermediate string (saving memory overhead), perhaps there's a way to have stdo of this command write straight to the files Writer interface.

pkg/harness/tool/v1/tool.go
kinjalh: since infracost is terraform specific, don't we want nil, nil returns for Ansible?
Similar to Plan(), State(), etc

michaeljguarino: Oh that's the right behavior i just didn't realize it would ensure this function is implemented for it

michaeljguarino (approved):
fix conflicts

michaeljguarino:
You should definitely regression test this on stacks w/o cost set up too

PR #751, opened 2026-05-06
feat: Add a download_manifests tool to our agent harness mcp server

JohnBlackwell:
This should:

- take a cluster handle + service name and fetch the manifests using our service files gql api.

- place those files in a separate subdirectory (use the {handle}-{name} syntax to make it semantically meaningful perhaps)

- instruct the agent harness that the files are located there and it can inspect them as it wishes

This will help agents better understand both plural's gitops system + things like external charts without mistakenly guessing via web searches.

## Test Plan

Test environment: https://jbdemo.console.cloud.plural.sh/

agent run -> https://jbdemo.console.cloud.plural.sh/ai/agent-runs/cd0b7812-4d50-494a-ae1f-f8806155382d

## Checklist

    - [x] Sync waves annotations are respected.
    - [x] Sync phases annotations are respected. Phases are executed in the correct order.
    - [x] Sync hook delete policies are respected. Resources are not recreated once they reach the desired state.
    - [x] Agent logs do not contain any errors (after running for at least 30 minutes).
    - [x] There are no visible anomalies in Datadog (after running for at least 30 minutes).


internal/mcpserver/agent/tool/downloadmanifests.go
```diff
@@ -0,0 +1,174 @@
...
+
+	svc, err := in.client.GetServiceDeploymentByHandle(in.ClusterHandle, in.ServiceName)
+	if err != nil {
+		return mcp.NewToolResultError(fmt.Sprintf("failed to look up service %q on cluster %q: %v", in.ServiceName, in.ClusterHandle, err)), nil
+	}
+	if svc.Tarball == nil || *svc.Tarball == "" {
+		return mcp.NewToolResultError(fmt.Sprintf("service %q on cluster %q does not yet have a rendered tarball available", in.ServiceName, in.ClusterHandle)), nil
+	}
+
+	_, token := in.client.GetCredentials()
```
michaeljguarino: I'm pretty sure this is a user token not deploy token, which cannot use the tarball url.  If it can, this is fine but should confirm.

If it's usign a deploy token, it needs to be switched to a user token though.

michaeljguarino (commented):
also I'm pretty sure the agent harness pods have read-only root filesystem enabled, we should make sure it's writing to an emptyDir volume that is allowed for r/w access (or add another so we can)

JohnBlackwell:
@greptile

JohnBlackwell:
@greptile

JohnBlackwell:
@greptile

michaeljguarino (approved):

